PR #8, opened 2022-05-12
Rename on*Filter functions to on*Mapper

reshadi:
the onArgsFilter and onValueFilter functions were not really filtering
values, rather they were mapping input values to new values.
This method rename makes them more meaningful.


reshadi:
This is already done, no longer need the PR

PR #16, opened 2023-01-26
Intercepting react api

reshadi:
This is generic code to enable intercepting react api.
(Most of the code if migrated from Flowjs internal repo to here)


packages/hyperion-autologging/src/Surface.ts
```diff
@@ -0,0 +1,341 @@
...
+   * sure that is not added to the DOM tree.
+   * The following code blocks the effect of `setAttribute` by returning true
+   */
+  IElement.setAttribute.onArgsObserverAdd(function (
+    this: Element,
+    attrName,
+    attrValue,
+  ) {
+    if (
+      attrName === domSurfaceAttributeName &&
```
schwartzmx, line 153: should we consider the `SurfaceDOMString.toString() !== ''`?

packages/hyperion-autologging/src/SurfaceConsts.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+ */
+
+'use strict';
+
+export const AUTO_LOGGING_SURFACE = "data-surface";
```
schwartzmx, line 7: intentionally leaving off `id` suffix?

packages/hyperion-autologging/src/SurfaceProxy.ts
```diff
@@ -0,0 +1,73 @@
...
+   * separate container DOM node and shown in place later.
+   * Although DOM tree hierarchy is broken, the React Context hierarchy will
+   * continue to work. So, we use that fact and assign the right surface attribute
+   * to the container node. This way, when we walk up the DOM tree, we find the
+   * right surface value.
+   */
+  IReactDOMModule.createPortal.onArgsMapperAdd(args => {
+    const [_node, container] = args;
+    if (container instanceof HTMLElement) {
+      // args[0] = options.ReactModule.createElement(SurfaceProxy, { container }, node);
```
schwartzmx, line 64: could we bundle the changes here where we found the portal body container surface attribute mutating?

packages/hyperion-react/src/IReactComponent.ts
```diff
@@ -0,0 +1,303 @@
...
+  function processReactFunctionComponent(functionComponent: TReactFunctionComponent): TReactFunctionComponent {
+    /**
+     * For functional components, we should always replace them with the
+     * intercepted version of them.
+     * however, the interceptFunction itself will only assign a FunctionIntercetpr
+     * to the function once.
+     * So, we don't need to use the interceptionInfo map here.
+     */
+
+    const fi = __DEV__
```
schwartzmx, line 157: will `__DEV__` work as expected here?

packages/hyperion-autologging/src/SurfaceConsts.ts
reshadi: Yes, we should have done that long ago, I put the default without `id' but of course can be changed with a config.

packages/hyperion-autologging/src/SurfaceProxy.ts
reshadi: I will have to pull in your fix from WWW and directly continue using the <SurfaceProxy>

packages/hyperion-react/src/IReactComponent.ts
reshadi: Yes, I made sure of that, it works both separately if you use the hyperion-test-react-app and also when we pull in the code into WWW

schwartzmx (approved):
overall looks good,  accepting to unblock.  I think we know there are still a couple changes to port over.

vsylee (approved):

PR #20, opened 2023-02-14
Transfer createPortal interception changes

schwartzmx:
Note: still a bit new to Typescript and felt I had to do a fair amount of type gymnastics to get it to work.   The Flowlet types have to be propagated across a lot of types/options.

There also is a bit of a circular dependency on ALSurface and ALSurfaceProxy now because the proxy is now wrapping using Surface component which is nested in `init` within ALSurface.  So I pulled out some of the shared types into Types.ts (which seems to be a pattern we use).

Let me know what your thoughts are on this,  @reshadi - or if you have a better way to handle this I'm all ears.


reshadi (changes requested):
Please rebase on the latest changes that has mostly removed the type parameters. It should be now easier to apply your changes.

schwartzmx:
Updated,  still not sure if it's the cleanest way.
Since we are adding flowlet to SurfaceContext we need to propagate the types,  and also since we are re-wrapping in Surface component we also need the same types used in the underlying surface using portal when it's intercepted.

packages/hyperion-autologging/src/ALSurfaceContext.ts
```diff
@@ -17,37 +20,55 @@ export type InitOptions =
...
+  FlowletType extends Flowlet<DataType>
+>(): ALSurfaceContextValue<DataType, FlowletType> {
+  return {
+    surface: '',
+    flowlet: null,
+  };
 };
 
-export let ALSurfaceContext: React.Context<ALSurfaceContextValue> | null = null;
+export let ALSurfaceContext: React.Context<any> | null = null;
```
reshadi, line 43: why 'any'? I assume that's because of the type parameters that you added, which tells you why it is not a good idea to do that and then break the type safety somewhere else.

packages/hyperion-autologging/src/ALSurfaceContext.ts
```diff
@@ -17,37 +20,55 @@ export type InitOptions =
...
 
-const DefaultSurfaceContext: ALSurfaceContextValue = {
-  surface: '',
+
+
+function getDefaultSurfaceContext<
+  DataType extends ALFlowletDataType,
+  FlowletType extends Flowlet<DataType>
+>(): ALSurfaceContextValue<DataType, FlowletType> {
+  return {
```
reshadi, line 37: Don't create a default everytime. keep the constant.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -5,10 +5,13 @@
 'use strict';
 
 
-import type * as React from 'react';
+import * as React from 'react';
```
reshadi, line 8: Must keep type here.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -40,34 +43,61 @@ import * as IReactDOM from "@hyperion/hyperion-react/src/IReactDOM";
...
-    domSurfaceAttributeName?: string;
+    surfaceComponent?: SurfaceComponent<DataType, FlowletType, FlowletManagerType>;
+    flowletManager: FlowletManagerType;
   }>;
 
-export function init(options: InitOptions): void {
-  const { domSurfaceAttributeName = AUTO_LOGGING_SURFACE, IReactDOMModule } = options;
-
+export function init<
+  DataType extends ALFlowletDataType,
```
reshadi, line 58: remove all of this, you should not need any of this

packages/hyperion-autologging/src/ALSurfaceContext.ts
```diff
@@ -17,37 +20,55 @@ export type InitOptions =
   }>;
 
 
-type ALSurfaceContextValue = Readonly<{
+type ALSurfaceContextValue<
+  DataType extends ALFlowletDataType,
+  FlowletType extends Flowlet<DataType>
+> = Readonly<{
   surface: string,
+  flowlet: FlowletType | null,
```
reshadi, line 28: Just to make sure these values are always assigned together, you should define the type like the following
`{ surface: string; flowlet: FlowletType} | {surface: null; flowlet: null}`

TS is smart enough to not allow a broken combination. Then the code bellow becomes easier.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -40,34 +43,61 @@ import * as IReactDOM from "@hyperion/hyperion-react/src/IReactDOM";
...
+    const [node, _container] = args;
+
+    if (node != null && React.isValidElement(node)) {
+      const { surface, flowlet: surfaceFlowlet } = useALSurfaceContext<DataType, FlowletType>();
       if (surface != null) {
-        container.setAttribute(domSurfaceAttributeName, surface);
+        // If the surface of the flowlet is the same surface as the one we are proxy wrapping
+        // because of portal, pass that to this wrapping surface component.
+        let flowlet = surfaceFlowlet;
+        if (flowlet == null) {
```
reshadi, line 78: see above. If you fix the type, it would be guaranteed that flowlet much have a value here. You can also `assert` that it is not null here and avoid the complex (and possibly wrong) recalculation of flowlet.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -40,34 +43,61 @@ import * as IReactDOM from "@hyperion/hyperion-react/src/IReactDOM";
...
+          const topFlowlet = flowletManager.top();
+          if (topFlowlet == null || topFlowlet?.name !== surface) {
+            flowlet = flowletManager.push(new flowletManager.flowletCtor(surface, topFlowlet));
+          } else {
+            flowlet = topFlowlet;
+          }
+        }
+        if (options.surfaceComponent) {
```
reshadi, line 86: why should it be optional?

packages/hyperion-autologging/src/Types.ts
```diff
@@ -0,0 +1,31 @@
...
+  FlowletManagerType extends FlowletManager<FlowletType>> = (props: IReactPropsExtension.ExtendedProps<SurfacePropsExtension<DataType, FlowletType>> & {
+    flowlet: FlowletType,
+    flowletManager: FlowletManagerType,
+    fullSurfaceString?: string
+  }
+  ) => React.ReactElement;
+
+
+
+export class SurfacePropsExtension<
```
reshadi, line 25: Such files will grow and age very badly. Make this class into a proper file.

reshadi (changes requested):

packages/hyperion-autologging/src/ALSurfaceProxy.ts
schwartzmx: `Surface.init` would require it then,  which is not possible to provide as is because it's nested within.   Is there a way to have it's init options (`SurfaceProxy`) required except for `surfaceComponent` and only require it via `SurfaceProxy.init`?  Would that warrant two separate types,  or is there a trick to require it only when `SurfaceProxy.init` is called but still share the init requirements for when Surface.init is called?

schwartzmx: Updated in latest push to use two types,  but if there's a better way I can change it

reshadi (approved):

PR #21, opened 2023-02-14
Remove type params from the init functions

reshadi:
Making it easier to apply createPortal fix


schwartzmx:
Thanks for this,  I think it will make what I was doing in #20 much simpler.   Looks good to me.

schwartzmx (approved):

PR #22, opened 2023-02-21
Intercept XHR response

vsylee:
Added a new hook for the XHR `loadend` event, also modifying the existing `onNetworkRequest` hook to include the XHR request instance itself in order to enable tracking the same flowlet with the request instance. Requesting feedback on this approach and how to best scale this for `fetch` calls as well.

### DIFF
--- a/packages/hyperion-util/src/onNetworkRequest.ts
+++ b/packages/hyperion-util/src/onNetworkRequest.ts
@@ -3,15 +3,15 @@
  */
 
 import { Hook } from "@hyperion/hook";
-import * as IWindow from "@hyperion/hyperion-dom/src/IWindow";
+// import * as IWindow from "@hyperion/hyperion-dom/src/IWindow";
 import * as IXMLHttpRequest from "@hyperion/hyperion-dom/src/IXMLHttpRequest";
 import * as intercept from "@hyperion/hyperion-core/src/intercept";
 import { assert } from "@hyperion/global";
 
-type NetworkRequest = {
+type NetworkRequestInfo = {
   readonly body?: RequestInit['body'],
   readonly method: string;
-  readonly url: string | URL; // TODO: should we always send on of the types?
+  readonly url: string;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/hyperion-util/src/onNetworkRequest.ts
```diff
@@ -3,58 +3,80 @@
...
-  }
+export const onNetworkRequestSend = new Hook<(request: XMLHttpRequest, requestInfo: NetworkRequestInfo) => void>();
 
-  onNetworkRequest.call(request);
-});
+/**
+ * This is a generic event to be fired when a network request has completed,
+ * whether successfully or unsuccessfully.
+ */
+export const onNetworkRequestLoad = new Hook<(request: XMLHttpRequest) => void>();
```
reshadi, line 28: why not keep these as 
`onNetworkRequest` & `onNetworkResponse`?

packages/hyperion-util/src/onNetworkRequest.ts
```diff
@@ -3,58 +3,80 @@
...
-      method: input.method,
-      url: input.url,
-    }
-  } else {
-    request = {
-      method: "GET",
-      url: input
-    }
-  }
+export const onNetworkRequestSend = new Hook<(request: XMLHttpRequest, requestInfo: NetworkRequestInfo) => void>();
```
reshadi, line 22: If we are making everything very XHR specific, then do we even need to have yet another set of events for "onNetwork` defined here? Wouldn't it be easier to just expose XHR?

Also, alternatively, I think we can just add an `initiatorXHR?: XMLHttpRequest` to the `NetworkRequestInfo` itself and continue keeping it generic.

packages/hyperion-util/src/onNetworkRequest.ts
```diff
@@ -3,58 +3,80 @@
...
+   });
 });
 
 IXMLHttpRequest.send.onArgsObserverAdd(function (this, body) {
-  const request = intercept.getVirtualPropertyValue<NetworkRequest>(this, XHR_REQUEST_INFO_PROP);
-  assert(request != null, `Unexpected situation! Request info is missing from xhr object`);
-  onNetworkRequest.call((body instanceof Document) ? request : {...request, body}); // assert already ensures request is not undefined
+  this.addEventListener('loadend', (event) => {
+    if (event.target instanceof XMLHttpRequest) {
+      onNetworkRequestLoad.call(event.target);
```
reshadi, line 69: Could we just read the response and send it in the event instead of the object itself?

packages/hyperion-util/src/onNetworkRequest.ts
```diff
@@ -3,58 +3,80 @@
...
+    }
+  });
```
reshadi, line 71: Just to be sure, add a `once` to the `addEventListener`

reshadi:
Already incorporated in the `ALNetworkPublisher` of `hyperion-autologging` package.

PR #24, opened 2023-02-25
making sure useSurfaceContext is called inside of a React component.

reshadi:
Needed to bring back the `SurfaceProxy` component to wrap the node in `createPortal` and avoid calling the `useContext` hook outside of a React component. otherwise, the application will be flooded by errors.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceProxy.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceProxy.ts
@@ -13,35 +13,6 @@ import { FlowletManagerType, SurfaceComponent } from './ALSurface';
 import { assert } from '@hyperion/global';
 
 
-/**
- * We need to use a hook to get the surface value, but the rules of using
- * hooks won't allow us to call it outside of a react component.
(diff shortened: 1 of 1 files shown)


schwartzmx (approved):
lgtm,  thanks!

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -56,8 +27,33 @@ type ProxyInitOptions =
...
 
+/**
+ * We need to use a hook to get the surface value, but the rules of using
+ * hooks won't allow us to call it outside of a react component.
+ * So, the following proxy component is purely for getting around this
+ * limitation and reading the current surface value during rendering.
+ * If we can find a way around this limitation, we can use a simpler logic
+ * like the following:
+ */
+function SurfaceProxy(props: React.PropsWithChildren<ProxyInitOptions>) {
```
schwartzmx, line 38: Was thinking too,  even though we have to wrap in another component it might be helpful as it will show these cases in the component stack more clearly instead of just another `Surface`.

PR #27, opened 2023-02-28
Only wrap in Surface component if surface can be extracted from context when using proxy

schwartzmx:
The assertions were still letting through constructions of surfaceComponent,  with a passed null flowlet.  Causing a run time error accessing `name` from passed flowlet.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceProxy.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceProxy.ts
@@ -10,12 +10,11 @@ import { useALSurfaceContext } from './ALSurfaceContext';
 import * as IReactDOM from "@hyperion/hyperion-react/src/IReactDOM";
 import { SurfacePropsExtension } from './ALSurfacePropsExtension';
 import { FlowletManagerType, SurfaceComponent } from './ALSurface';
-import { assert } from '@hyperion/global';
 
 
 export type InitOptions =
   Readonly<{
-    ReactModule: { createElement: typeof React.createElement };
+    ReactModule: { createElement: typeof React.createElement, Fragment: typeof React.Fragment };
     IReactDOMModule: IReactDOM.IReactDOMModuleExports;
     flowletManager: FlowletManagerType;
   }>;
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -35,21 +34,23 @@ type ProxyInitOptions =
  * If we can find a way around this limitation, we can use a simpler logic
  * like the following:
  */
-function SurfaceProxy(props: React.PropsWithChildren<ProxyInitOptions>) {
+function SurfaceProxy(props: React.PropsWithChildren<ProxyInitOptions>): React.ReactElement {
```
reshadi, line 37: you can return `React.Node` here, which is the same type as `props.children`.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -35,21 +34,23 @@ type ProxyInitOptions =
...
+      {
+        __ext: new SurfacePropsExtension(flowlet),
+        flowlet: flowlet,
+        flowletManager: flowletManager,
+        fullSurfaceString: surface,
+      },
+      children
+    );
+  } else {
+    return ReactModule.createElement('div', props, children);
```
reshadi, line 52: ... and then here just return `children`.

reshadi (commented):
Thanks for debugging it. You can simplify it further.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
schwartzmx: Typescript was complaining even before the return type was added.    I tried returning just `children` but children can be undefined (even though we check node != null),  which causes a type error in the interception `createElement`.

```
type React.ReactNode = string | number | boolean | React.ReactElement<any, string | React.JSXElementConstructor<any>> | React.ReactFragment | React.ReactPortal | null | undefined
```
image

```
Type 'undefined' is not assignable to type 'ReactElement<any, any> | null'.ts(2769)
```

schwartzmx: What I just pushed won't work either,  getting `throwInvalidHookError`.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -35,25 +34,27 @@ type ProxyInitOptions =
  * If we can find a way around this limitation, we can use a simpler logic
  * like the following:
  */
-function SurfaceProxy(props: React.PropsWithChildren<ProxyInitOptions>) {
+function SurfaceProxy(props: React.PropsWithChildren<ProxyInitOptions>): React.ReactElement | React.ReactNode {
```
reshadi, line 37: You can instead use the following: 
`ReactModule.createElement(React.Fragment, {}, children);`

That is the same as using `<>{childern}</>` in the jsx files.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
```diff
@@ -65,7 +66,12 @@ export function init(options: ProxyInitOptions): void {
     const [node, _container] = args;
 
     if (node != null) {
-      args[0] = ReactModule.createElement(SurfaceProxy, options, node);
+      const proxy = SurfaceProxy({...options, children: node});
```
reshadi, line 69: this is going to cause problem with the hook again and you said.

packages/hyperion-autologging/src/ALSurfaceProxy.ts
schwartzmx: Yeah this is what I was just starting to look into,  thanks for the tip!

PR #28, opened 2023-03-01
Performance absolute now

reshadi:


reshadi:
A more complete one is sent separately.

PR #29, opened 2023-03-01
Heartbeat

reshadi:
The is a collection of a few fixes on the old code, enabling the heartbeat, and and updating the autologging init function.


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -153,26 +166,22 @@ export function publish(
       autoLoggingID: getOrSetAutoLoggingID(element),
       flowlet,
       isTrusted,
+      /**
+       * eventIndex will be overridden in the subscriber, we are passing in a value here because
+       * Flow requires it but because some al_ui_events will be filtered out before emitting
+       * events to Falco, we can't generate the event_index here
+       */
       eventIndex: 0,
```
schwartzmx, line 174: I know this is already here but in what cases does this happen?  Seems odd to make the subscriber generate it,  technically it happened regardless of if it gets filtered later downstream or not.

schwartzmx (approved):
overall looks good

will have the mutation publisher following this getting merged

packages/hyperion-autologging/src/ALHeartbeat.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export type InitOptions = Types.Options<{
+  channel: ALHeartbeatChannel;
+  heartbeatInterval?: number;
+  maxUserInactivityDuration?: number;
+}>;
+
+const HEARTBEAT_INTERVAL = 30 * 1000 /* DateConsts.MS_PER_SEC */;
+const MAX_USER_INACTIVITY_DURATION = 4 * HEARTBEAT_INTERVAL;
+const VISIBILITY_CHANE_EVENT = "visibilitychange";
```
vsylee, line 42:
```suggestion
const VISIBILITY_CHANGE_EVENT = "visibilitychange";
```

packages/hyperion-autologging/src/ALHeartbeat.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const userActionListener = channel.addListener(
+    'al_ui_event',
+    ({ eventTimestamp }) => {
+      _lastUserActionTime = eventTimestamp;
+    },
+  );
+
+  const pageVisibilityListener = (_e: Event) => {
+    const isHidden = document.hidden;
```
vsylee, line 82: WWW also checks for `(moz|ms|webkit)Hidden` (referencing `html/shared/core/Visibility.js`) - should we be doing the same here?

reshadi: Yes, I checked it but that seems to be pretty old. All modern browsers do support this already .

https://caniuse.com/?search=visibilitychange
https://caniuse.com/mdn-api_document_hidden

That's why I decided to with a simpler code.

PR #30, opened 2023-03-03
Surface Mutation Publisher

schwartzmx:
Added initial version of surface mutation publisher
- Some things to note:
   - In the original version we have a reliance on  (ads-specific) interaction callbacks (onAdd, onEnd) ending before we would process the mutation events,  see `schedulePostInteractionTask` and `onInteraction` in the old code. 
     - How should we handle this,  should we just arbitrarily choose a timeout and use `TimedTrigger`?  The initial PR is just firing them immediately and does not have a sense of added/removed/active.  Replaced with only `active`.
       
Included caching react element info for both the mutation publisher, and UI event publisher,  which is configurable via initoptions,  this is important as this is one of the first things fixed by IFB team when they first used the generic UI publisher.
- Some things to note:
   - The initial version is about as simple as it gets,  rather than port all the `ElementInfo`/`UIState` (virtualProperty setting etc.).  I just went with the basics and grab the info and store it in the initial event,  caching it, and then it is available when ultimately fired.
   
Also included some small fixes with surface attribute passing,  added to UI publisher, removed some hard reliance on the const and instead use init option specified surface attribute.
- We seem to be having some shared attributes across configuration,  maybe we could pull this into the top level configuration and pass it down along with feature-specific options.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
@@ -7,7 +7,7 @@
 import type { ALChannelSurfaceEvent } from './ALSurface';
 import type { Channel } from "@hyperion/hook/src/Channel";
 import * as Types from "@hyperion/hyperion-util/src/Types";
-import { ALLoggableEvent } from "./ALType";
+import { ALLoggableEvent, ALReactElementEvent } from "./ALType";
 
 import { ALFlowlet, ALFlowletManager } from "./ALFlowletManager";
 import * as ALID from './ALID';
@@ -17,14 +17,12 @@ import { ComponentNameValidator, getReactComponentData_THIS_CAN_BREAK, ReactComp
 import { AUTO_LOGGING_SURFACE } from './ALSurfaceConsts';
 import { getElementName } from './ALInteractableDOMElement';
 
-type ALMutationEvent =  Readonly<{
+type ALMutationEvent = ALReactElementEvent & Readonly<{
   event: 'mount_component' | 'unmount_component';
   element: HTMLElement,
   elementName: string | null,
   mountedDuration?: number;
   flowlet?: ALFlowlet | null;
-  reactComponentName?: string | null,
-  reactComponentStack?: string[] | null,
 }>;
 
 export type AdsALSurfaceMutationEventData = Readonly<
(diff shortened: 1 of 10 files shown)


packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -12,11 +12,32 @@ import { ALLoggableEvent } from "./ALType";
...
+import { getElementName } from './ALInteractableDOMElement';
 
+type ALMutationEvent =  Readonly<{
+  event: 'mount_component' | 'unmount_component';
+  element: HTMLElement,
+  elementName: string | null,
+  mountedDuration?: number;
+  flowlet?: ALFlowlet | null;
+  reactComponentName?: string | null,
+  reactComponentStack?: string[] | null,
```
reshadi, line 27: Did you want to use `ALReactElementEvent ` here?

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -105,14 +112,20 @@ export function publish(options: InitOptions): void {
       flowlet = new ALFlowlet(`ts${captureTimestamp}`, flowlet);
       flowlet = flowletManager.push(flowlet);
 
+      let reactComponentData: ReactComponentData | null = null;
+      if (cacheElementReactInfo) {
+        reactComponentData = getReactComponentData_THIS_CAN_BREAK(element, componentNameValidator);
```
reshadi, line 117: This is potentially the only part that can add some perf overhead, but I think for now we can let it happen this way and see what the perf impact is. it is possible that it all would depends on the frequency of these events.

reshadi (approved):
Thanks for diligently applying the changes. The concerns you raised are valid, but we can allow it for now and see if they really cause that much performance overhear. 

Did you check to see if all of this works find with the react-test-app?

reshadi:
> * We seem to be having some shared attributes across configuration,  maybe we could pull this into the top level configuration and pass it down along with feature-specific options.


I agree, in fact when I was importing this stuff back into WWW, I noticed the repetitive stuff and was thinking of making a shared part (e.g. channel, flowletManager, ...) Once all the options are added we can do another pass and factor shared options out. We can do that just in the very top init function and still keep each one of individual modules independent. Once you merge this code, I'll add that part and import in the WWW

schwartzmx:
Made type changes, added a component to toggle surface render to easily see mount/unmount.  Also TIL about React.StrictMode double mount,  I saw this in the console before and thought it was local dev related, but with toggle component it's more apparent the effects are called twice.

Locale console output from channel:
image

reshadi (approved):
Looks great. Chaning the event name is a good idea and makes it more clear.

PR #32, opened 2023-03-03
Update the AutoLogging init to handle more cases

reshadi:
Now the init can be called multiple time and after the first time it returns cached results.

This has multiple benefits:
- We can have some components eagerly init the AL. That is usefull if one shared component is used in multiple applications.
- We can init AL in one place, and the in other places just read the results, e.g. for initializing the SurfaceRenderer component.


packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -0,0 +1,79 @@
+/**
```
reshadi, line 1: this is actually the former `index.ts` and I used `git mv` to rename the file, not sure why GH shows it this way.

packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return cachedResults;
+}
+
+export function getSurfaceRenderer(): InitResults['surfaceRenderer'] {
+  const renderer = cachedResults?.surfaceRenderer;
+  assert(
+    renderer != null,
+    "AutoLogging must have been initilized first. Did you forget to call .init() functions?",
+    {
```
reshadi, line 68: Note this is a trick that both asserts locally, and also throws and exception for other apps to catch and report it!

schwartzmx (approved):
lgtm,  good change with being able to grab the render anywhere

PR #41, opened 2023-03-13
Add surface to al_ui_event

kahyoung:
[ALUIEventData looks like it should have surface provided](https://github.com/facebook/hyperion/blob/main/packages/hyperion-autologging/src/ALUIEventPublisher.ts#L51) but it doesn't currently do this.

I'm just passing it through from the ALUIEventCaptureData.


reshadi (approved):
Thanks for fixing this

reshadi:
@kahyoung  thanks for sending a fix. Looks like the compilation is failing. You need to convert the type of `surface` in the corresponding UIEvent into `string | null` or something like that to fix compilation problem.

PR #42, opened 2023-03-13
Adds basic ALElementInfo for caching element react info, and potentia…

schwartzmx:
…lly other expensive operations down the line

- Also included, propagate surface that is looked up in capture, but not emitted in ui event
- Add missing copyright
- Update usage of cache react info to use elementinfo
	- if some other operation needs access to the same element info it can be cached across event types with virtual prop,  or in the case of double clicking the same button we now will only incur one lookup for component info
	  - On the other hand, if the stack is updated we get stale info

### DIFF
--- a/packages/hyperion-autologging/src/ALElementInfo.ts
+++ b/packages/hyperion-autologging/src/ALElementInfo.ts
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+ */
+
+'use strict';
+
+import { ComponentNameValidator, defaultComponentNameValidator, getReactComponentData_THIS_CAN_BREAK } from './ALReactUtils';
+import type { ReactComponentData } from './ALReactUtils';
+import { getVirtualPropertyValue, setVirtualPropertyValue } from '@hyperion/hyperion-core/src/intercept';
+
+const AL_ELEMENT_INFO_PROPNAME = '__alInfo';
+const AUTO_LOGGING_COMPONENT_TYPE = 'data-auto-logging-component-type';
+
+export default class ALElementInfo {
+  element: Element;
+  private reactComponentData: ReactComponentData | null = null;
+  private reactComponentType: string | null = null;
+  private componentNameValidator: ComponentNameValidator = defaultComponentNameValidator;
+
+  constructor(
+    element: Element,
+    componentNameValidator?: ComponentNameValidator,
+  ) {
+    this.element = element;
+    setVirtualPropertyValue(
+      element,
+      AL_ELEMENT_INFO_PROPNAME,
+      this,
+    );
+
+    this.componentNameValidator = componentNameValidator ?? this.componentNameValidator;
+    this._cacheInfo();
+  }
+
+  _cacheInfo(): void {
+    this.getReactComponentData();
+    this.getReactComponentType();
+  }
+
+  static get(element: Element): ALElementInfo | undefined {
+    return getVirtualPropertyValue<ALElementInfo>(
+      element,
+      AL_ELEMENT_INFO_PROPNAME,
+    );
+  }
+
+  static getOrCreate(element: Element, componentNameValidator?: ComponentNameValidator): ALElementInfo {
+    return ALElementInfo.get(element) ?? new ALElementInfo(element, componentNameValidator);
+  }
+
+  getReactComponentData(): ReactComponentData | null {
+    if (!this.reactComponentData) {
+      this.reactComponentData = getReactComponentData_THIS_CAN_BREAK(
+        this.element,
+        this.componentNameValidator,
+      );
+    }
+    return this.reactComponentData;
+  }
+
+  getReactComponentType(): string | null {
+    if (!this.reactComponentType) {
+      this.reactComponentType = this.element.getAttribute(AUTO_LOGGING_COMPONENT_TYPE);
+    }
+    return this.reactComponentType;
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+import { ComponentNameValidator, defaultComponentNameValidator, getReactComponentData_THIS_CAN_BREAK } from './ALReactUtils';
+import type { ReactComponentData } from './ALReactUtils';
+import { getVirtualPropertyValue, setVirtualPropertyValue } from '@hyperion/hyperion-core/src/intercept';
+
+const AL_ELEMENT_INFO_PROPNAME = '__alInfo';
+const AUTO_LOGGING_COMPONENT_TYPE = 'data-auto-logging-component-type';
+
+export default class ALElementInfo {
+  element: Element;
+  _reactComponentData: ReactComponentData | null = null;
```
reshadi, line 16: in TS, there is no "convention" for private / public members of class. If you want these to be private, you should explicitly say that using "**private**" keyword.

packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+  _reactComponentType: string | null = null;
+  _componentNameValidator: ComponentNameValidator = defaultComponentNameValidator;
```
reshadi, line 18: Does this need to be stores per instance of this object? I assume this might be per app, and so may be it could be static?

packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (!this._reactComponentData) {
+      this._reactComponentData = getReactComponentData_THIS_CAN_BREAK(
+        this.element,
+        this._componentNameValidator,
+      );
+    }
+    return this._reactComponentData;
+  }
+
+  getReactComponentType(): string | null {
```
reshadi, line 61: are we using this anymore? I think in WWW that was left over from the previous implementation.

packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -58,7 +59,7 @@ export type InitOptions = Types.Options<{
 }>;
 
 export function publish(options: InitOptions): void {
-  const { domSurfaceAttributeName = AUTO_LOGGING_SURFACE, channel, flowletManager, cacheElementReactInfo, componentNameValidator = null } = options;
+  const { domSurfaceAttributeName = AUTO_LOGGING_SURFACE, channel, flowletManager, cacheElementReactInfo, componentNameValidator = defaultComponentNameValidator } = options;
```
reshadi, line 62: we assume `publish` is called once per app, so fair to say `componentNameValidator` is static. You can have a static `setComponentNameValidator` in info class and avoid reduce the size of that object.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -188,6 +190,7 @@ export function publish(options: InitOptions): void {
       autoLoggingID: getOrSetAutoLoggingID(element),
       flowlet,
       isTrusted,
+      surface,
```
reshadi, line 193: We just had another PR adding this. I'll approve and merge that to encourage more people to help! :)

reshadi (approved):
Overall looks good. only the static field conversion and private qualifiers need to be updated.

packages/hyperion-autologging/src/ALElementInfo.ts
schwartzmx: Yeah I think that makes sense,  thoughts on if I should create a parent class that underneath manages creating ALElementInfo instances,  where we can keep the object that is set via virtual property as slim as possible?

reshadi: Technically by using the  `static get(..)` method you achieved the same thing. If you wanted to have a parent class, then you had to create a singleton instance of that class to manage things. A (class + singleton) is almost the same as using (class + static) members on it.

packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.element = element;
+    setVirtualPropertyValue(
+      element,
+      AL_ELEMENT_INFO_PROPNAME,
+      this,
+    );
+    this._cacheInfo();
+  }
+
+  _cacheInfo(): void {
```
reshadi, line 42: private?

packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -0,0 +1,74 @@
...
+const AUTO_LOGGING_COMPONENT_TYPE = 'data-auto-logging-component-type';
+
+let componentNameValidator: ComponentNameValidator = defaultComponentNameValidator;
+/**
+ * Set a component validator function to use when extracting
+ * a component name for the event, utilized in ALElementInfo.
+ * The stack will remain unfiltered, but component name linked must be valid via validator.
+ * @param validator: callable passed a component name returning true if valid
+ */
+export function setComponentNameValidator(validator: ComponentNameValidator): void {
```
reshadi, line 21: technically, this could have been added to the `ALReactUtils` itself for `sedDefaultComponentNameValidator` so that any other module in the future can simply call the corresponding api without worrying about adding this stuff. 

I'll merge this commit, but you can fix it later.

reshadi (approved):

packages/hyperion-autologging/src/ALElementInfo.ts
schwartzmx: Makes sense.  I did just realize though,  that with this static approach whoever calls the API second (surface mutation publisher in current test app) will overwrite previous caller config (ui event publisher in test app).  I don't see a use case for the validators to be different,  but might be something to account for with shared init options so we only allow one value in the config, which gets passed down to modules that require them.

PR #45, opened 2023-03-16
Move set validator to ReactUtils

schwartzmx:


packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -13,10 +13,10 @@ import { ALFlowlet, ALFlowletManager } from "./ALFlowletManager";
 import * as ALID from './ALID';
 import * as ALEventIndex from './ALEventIndex';
 import performanceAbsoluteNow from '@hyperion/hyperion-util/src/performanceAbsoluteNow';
-import { ComponentNameValidator, defaultComponentNameValidator, ReactComponentData } from './ALReactUtils';
+import { ComponentNameValidator, defaultComponentNameValidator, ReactComponentData, setComponentNameValidator } from './ALReactUtils';
```
reshadi, line 16: I wonder if we should remove the option entirely from this module and ui event publisher, and have the `AutoLogging.ts` itself directly call `setComponentNameValidaor` once. 

that way we know that we cannot pass two different options for this.

reshadi (approved):

packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
schwartzmx: Yeah I think that might be more ideal,  although it will only be specific to events that deal with element-related events.  e.g. network,  heartbeat, etc won't require this.

schwartzmx: Yeah I think that might be more ideal,  although it will only be specific to events that deal with element-related events.  e.g. network,  heartbeat, etc won't require this.

PR #46, opened 2023-03-16
Make surface attribute option required in publisher

schwartzmx:


schwartzmx:
#49 includes this change with shared options

PR #48, opened 2023-03-16
Emit a surface mutation regardless of if an element is found via quer…

schwartzmx:
…yselector

Open to thoughts on this,  but given what we saw when surface attribute was missing it was bit hard to debug,  and made me think that perhaps we should still emit a mutation event even if the querySelector doesn't find the associated element that was rendered.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
@@ -21,7 +21,7 @@ import ALElementInfo from './ALElementInfo';
 type ALMutationEvent = ALReactElementEvent & Readonly<{
   event: 'mount_component' | 'unmount_component';
   surface: string;
-  element: HTMLElement;
+  element: HTMLElement | null;
   elementName: string | null;
   mountedDuration?: number;
   flowlet?: ALFlowlet | null;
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -56,33 +56,34 @@ export type InitOptions = Types.Options<{
...
 
-  function processNode(node: Node, action: 'added' | 'removed') {
+  function processNode(node: HTMLElement | null, surface: string, action: 'added' | 'removed') {
     const timestamp = performanceAbsoluteNow();
 
     const flowlet = flowletManager.top();
-    if (!(node instanceof HTMLElement) || /LINK|SCRIPT/.test(node.nodeName)) {
-      return;
-    }
-    const surface = node.getAttribute(domSurfaceAttributeName);
```
reshadi, line 75: Generally this is a good change, however, I'm worried that if we allow this kind of breakage to happen, then the rest of the application may break. For example, when we have UI event, we will look for enclosing surface for that target element and we would need to see that surface. If we don't catch the problem here, we might have a harder time to see what was problem somewhere else.

schwartzmx: True,  also I just remembered,  if you just want simple surface mount/unmount the `al_surface_mount` and `al_surface_unmount` events are also available which this change basically re-implements/supports (same events just lacking the element info).  Will close this.

PR #51, opened 2023-03-22
Alnetwork

reshadi:
Enable both network request `al_network_request` and network response `al_network_response` events from auto logging.


packages/hyperion-autologging/src/ALNetworkPublisher.ts
```diff
@@ -0,0 +1,190 @@
...
+    } else if (input instanceof Request) {
+      request = {
+        body: input.body,
+        method: input.method,
+        url: input.url,
+      };
+
+    } else {
+      request = {
+        method: "GET",
```
schwartzmx, line 93: We use `?? "get"` above, but `"GET"` here as fallback,  should we use the same casing or lower/upper all `method` values for consistency?

PR #52, opened 2023-03-23
 Incorporate the surface name into flowlet

reshadi:
To ensure that we always have a more meaningful name for flowlets, we incorporate the surface name in the flowlet so that we know (surface, uievent, alid, timestamp) for the flowlet.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -93,6 +93,7 @@ export function publish(options: InitOptions): void {
 
   let lastUIEvent: CurrentUIEvent | null;
   let flowlet: ALFlowlet;
+  const defaultTopFlowlet = new flowletManager.flowletCtor("/");
 
   newEventsToPublish.forEach(eventName => {
     // Track event in the capturing phase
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -103,20 +104,22 @@ export function publish(options: InitOptions): void {
...
-        const ALFlowlet = flowletManager.flowletCtor;
-        flowlet = new ALFlowlet(eventName, flowletManager.top());
-        flowlet = new ALFlowlet(autoLoggingID, flowlet);
-        flowlet = new ALFlowlet(`ts${captureTimestamp}`, flowlet);
+        if (surface) {
+          flowlet = flowlet.fork(surface);
+        }
+        flowlet = flowlet.fork(eventName).fork(autoLoggingID).fork(`ts${captureTimestamp}`);
         flowlet = flowletManager.push(flowlet);
+        console.log(flowlet.getFullName());
```
schwartzmx, line 122: Can you remove this log?

schwartzmx (approved):
Makes sense to me,  also the `fork` chaining seems cleaner and more clear

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi: oops! Sure. Thanks for catching it!

PR #55, opened 2023-03-23
Enable marking of network requests with flowlet name.

reshadi:
With this commit, one can pass a query param name `requestFlowletMarker` to enable the code that adds the query param with current flowlet name to the url of the request. That way, backend services can connect the requess to what was going on in the client at the time.

### DIFF
--- a/packages/hyperion-autologging/src/ALNetworkPublisher.ts
+++ b/packages/hyperion-autologging/src/ALNetworkPublisher.ts
@@ -65,12 +65,57 @@ export type InitOptions = Types.Options<
      * will generate request/response events. 
      */
     requestFilter?: (request: RequestInfo) => boolean;
+
+    /**
+     * If passed, it will enable marking all requests with 
+     * the value of current flowlet using the following as queryparam name
+     */
+    requestFlowletMarker?: string;
   }
 >;
 
+function urlAppendParam(url: string, paramName: string, paramValue: string): string {
+  // Note that it is not easy to use URL since we don't quite know the base of relative urls
+  const separator = url.indexOf("?") !== -1 ? "&" : "?";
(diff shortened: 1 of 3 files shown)


packages/hyperion-autologging/src/ALNetworkPublisher.ts
```diff
@@ -65,12 +65,57 @@ export type InitOptions = Types.Options<
...
+
+    /**
+     * If passed, it will enable marking all requests with 
+     * the value of current flowlet using the following as queryparam name
+     */
+    requestFlowletMarker?: string;
   }
 >;
 
+function urlAppendParam(url: string, paramName: string, paramValue: string): string {
```
vsylee, line 77: Thoughts on exposing this as a hook? We also currently append app-specific values like QPL flow instance ids, ad account id, etc in multiple places for various network requests. I was thinking we could use Hyperion to intercept in one place and simplify the existing setup.

vsylee (approved):

packages/hyperion-autologging/src/ALNetworkPublisher.ts
reshadi: I thought it as well, but seems like many of the other cases depend on the request, for example, if we later add support script/image/... for network, then we won't want to add some of the other info right? 

Also, what should be signature of the hook? For example, we can assume something like the following?

```
requestMapper?: (request: RequestInfo, searchParamsAppend: (paramName: string, paramValue: string)=>void)=>void
```

Then one can simply check the request, and call the `searchParamsAppend` with desired values. That should be flexible enough for all use cases, right?

vsylee (approved):
Super excited for this, thanks!

PR #57, opened 2023-03-27
Support for key press/down events

schwartzmx:
- Uses generic ui publisher with some minor adjustments/changes
- Config whether to limit to interactable
- Emit dom event in UI events for subscribers to introspect/access additional properties
- Use set of UI events so that configuration can be handled on an event by event basis (up for debate on if this route makes sense)
- Add eventFilter, allows us to pre-filter events very early, for example specific key presses or specific elements we want to only track.

image


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -16,10 +16,12 @@ import { getSurfacePath } from "./ALSurfaceUtils";
 import { ALFlowletEvent, ALReactElementEvent, ALSharedInitOptions, ALTimedEvent } from "./ALType";
 
 export type ALUIEvent = Readonly<{
-  event: string,
-  element: Element,
+  domEvent: GlobalEventHandlersEventMap[ALUIEvent['event']],
```
reshadi, line 19: This might slowdown the compiler. Why not just `keyof GlobalEventHandlersEventMap`?

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -72,67 +74,92 @@ type CurrentUIEvent = {
 const PUBLISHED_EVENTS = new Set<string>();
 type ALChannel = Channel<ALChannelUIEvent>;
 
+interface UIEventConfig {
```
reshadi, line 77: I can see this might be handy for fine grained control of events.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -146,14 +173,34 @@ export function publish(options: InitOptions): void {
...
+          element = event.target instanceof HTMLElement ? event.target : null;
+        }
+
         if (element != null) {
           channel.emit('al_ui_event_bubble', {
+            domEvent: event,
             event: eventName,
             element,
-            bubbleTimestamp: performanceAbsoluteNow(),
+            bubbleTimestamp: bubbleTimestamp,
```
reshadi, line 201: name and value are the same, you can just use the name.

packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -13,15 +13,15 @@ import { ComponentNameValidator, setComponentNameValidator } from "./ALReactUtil
 import * as ALSurface from "./ALSurface";
 import * as ALSurfaceMutationPublisher from "./ALSurfaceMutationPublisher";
 import { ALSharedInitOptions } from "./ALType";
-import * as ALUIeventPublisher from "./ALUIEventPublisher";
+import * as ALUIEventPublisher from "./ALUIEventPublisher";
```
reshadi, line 16: 😲 👍

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -72,67 +74,92 @@ type CurrentUIEvent = {
...
   let flowlet: ALFlowlet;
   const defaultTopFlowlet = new flowletManager.flowletCtor("/");
 
-  newEventsToPublish.forEach(eventName => {
+  newEventsToPublish.forEach((eventConfig => {
+    const {eventName, eventFilter, cacheElementReactInfo, interactableElementsOnly = true } = eventConfig;
     // Track event in the capturing phase
-    window.document.addEventListener(eventName, (event: Event) => {
-      const element = getInteractable(event.target, eventName);
+    window.document.addEventListener(eventName, (event: GlobalEventHandlersEventMap[typeof eventConfig.eventName]) => {
```
reshadi, line 110: the type of `event` won't further narrow down because `eventName` is string. That kind of thing only works if you have generics. You might as well just keep it as `MouseEvent | KeyboardEvent`

reshadi (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
schwartzmx: I was hoping there was a way to narrow down to the specific event type based on the actual event in the object but I think it's a bit too clever.  e.g. If event = `keyup` the domEvent would be `KeyboardEvent` typed.

reshadi: there is that possibility if you break this type into a bunch of unions for example:
```
{
event: 'click' | 'mouseover' | 'mousedown'; 
domEvent: GlobalEventHandlersEventMap[ALUIEvent['event']]; // or just MouseEvent
} | {
event: 'keydown' | 'keyup' | 'keypress'; 
domEvent: GlobalEventHandlersEventMap[ALUIEvent['event']]; // or just KeyboardEvent
} | {
event: string;
domEvent: GlobalEventHandlersEventMap[ALUIEvent['event']]; // or just Event
}
```

schwartzmx: Oh good idea.  Even though it's a bit tedious and ugly I think it will be more clean on the subscriber's end to not have to do instance of assertions every place they want to inspect the event attributes.

schwartzmx: This actually gets me very close to what I want but when I try to use `eventFilter` later on in the listener it wants to account for every type of event
```ts
type UIEventConfigMapping<T> = {[K in keyof T]: {
  readonly eventName: K,
  // A callable filter for this event, returning true if the event should be emitted, or false if it should be discarded up front
  eventFilter?: (domEvent: T[K]) => boolean;
  // Whether to limit to elements that are "interactable", e.g. have event handlers registered to the element.  Defaults to true.
  interactableElementsOnly?: boolean;
  // Whether to cache element's react information on capture, defaults to false.
  cacheElementReactInfo?: boolean;
}}[keyof T];

type UIEventConfig = UIEventConfigMapping<GlobalEventHandlersEventMap>;
```
image

Is manually unioning the permutations the only way to solve that?

reshadi: no, union is not the only way, the way you're approaching it is actually good (better). Assuming you do trust the `document.addEventListener` to give you the right type, you may have to cast the step to get around the compile error. That might be a reasonable compromise in this case.

schwartzmx: Pushed up a new commit that "better" types the ALUIEvent,  so that in subscribers the domEvent can be narrowed by checking event is X (`click`,  `keydown` etc...).   Also included this change in the UIEventConfig, so that in configuration the type is stricter for the filter function defined.

All this at the expense of a couple `as any` casts I seemingly (might be a way around this) have to do to satisfy the compiler when we are looping through event config and processing the union of dom event types.  Let me know if you think this is better or worse,  I know there's tradeoffs but maybe I missed something.   I think this makes it a bit easier for usage in AutoLogging.init and also in the subscribers to al_ui_event.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -155,7 +161,7 @@ export function publish(options: InitOptions): void {
       }
       const eventData: ALUIEventCaptureData = {
         domEvent: event,
-        event: eventName,
+        event: (eventName as any),
```
reshadi, line 164: would the following work?
`eventName as ALUIEvent['event']`

Slightly better than any.

packages/hyperion-react-testapp/src/AutoLoggingWrapper.ts
```diff
@@ -47,18 +47,19 @@ export function init() {
...
           eventName: 'click',
           cacheElementReactInfo: true,
+          eventFilter: (domEvent: MouseEvent) => domEvent.isTrusted
         },
         {
           eventName: 'keydown',
           cacheElementReactInfo: true,
           interactableElementsOnly: true,
-          eventFilter: (domEvent: Event) => domEvent instanceof KeyboardEvent && domEvent?.code === 'Enter',
+          eventFilter: (domEvent: KeyboardEvent) => domEvent?.code === 'Enter',
```
reshadi, line 56: do you still need to specify the `KeyboardEvent`, or will it infer it correctly?

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -15,14 +15,15 @@ import { ReactComponentData } from "./ALReactUtils";
...
 
-export type ALUIEvent = Readonly<{
-  domEvent: GlobalEventHandlersEventMap[ALUIEvent['event']],
-  event: keyof GlobalEventHandlersEventMap,
-  element: Element | null,
-  elementName?: string | null,
-  isTrusted: boolean,
-
-}>;
+export type ALUIEvent<T = EventHandlerMap> = {
```
reshadi, line 18: Took me a while to understand this is not a nested object. May be either add some more comments, or break this into a `ALUIEventMap` and then define `ALUIEvent` accordingly. 

Also, I'm not sure you need to or should export this as a generic type. We don't expect external modules to change these types.

reshadi (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
schwartzmx: Doesn't seem to work,  then it says `domEvent` is incompatible
image

packages/hyperion-react-testapp/src/AutoLoggingWrapper.ts
schwartzmx: Nope it's not needed, it infers correctly which is nice.
image

schwartzmx:
@reshadi could you please re-review the last couple commits
- remove type exports,  add comments on what they are doing i.e. generating a union of event types
- Noticed in some cases with multiple UI events,  the single flowlet ref being held could cause interspersed event flowlets to get stuck (and not popped),  utilize a flowlet map by event name
- bubble event should be emitted even if element is null, since now we won't enforce interactable across all events unless specified,  the logic was changed to look at `event.target` instead of element

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -118,7 +118,7 @@ export function publish(options: InitOptions): void {
   trackInteractable(newEventsToPublish.map(ev => ev.eventName));
 
   let lastUIEvent: CurrentUIEvent | null;
-  let flowlet: ALFlowlet;
+  let flowletMap = new Map<UIEventConfig['eventName'], ALFlowlet>();
```
reshadi, line 135: Using map is one option. The `FlowletManager.pop` itself searches within the active flowlets for a matching flowlet to pop. That's why I required an argument in `.pop`.  The main problem is the "shared" flowlet variable between calls. Separating between calls by eventName should work most of the time, but just know that one other option is hang is on the target element itself. That the strategy I used for both XHR and Fetch in the `ALNetworkPublisher.ts`.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -232,16 +233,17 @@ export function publish(options: InitOptions): void {
             }
           }
         }
-        if (event.isTrusted) {
-          flowletManager.pop(flowlet);
+        if (event.isTrusted && flowletMap.has(eventName)) {
```
reshadi: Do you need the `isTrusted` now again?  I noticed that condition above has also `event.bubbles`. So, we can keep that list of conditions in one place, and only check the map here. 

Also, `.has` and `.get` have similar complexity. So, better to call `.get` once and then check null instead of `.has`

reshadi (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
schwartzmx: I was thinking there could be a case where we are tracking multiple event types that could be occurring on the same target,  e.g. mousemove and click on the same element.

reshadi (approved):

PR #58, opened 2023-03-28
Fix visibility issue in jest, separate react dom & comp prop extension

reshadi:
from each other.

In cases that we have to add surface wrappers (<span>) elements to the DOM, JEST may get confuse the wrapper with the compoennt itself and some assertions may fail.
This commit ensures that jest related attributes cannot be added to the wrappers and they stay invisible to the framework. (Technically, we could just prevent any non-verified attributes on these wrappers, but that might be too drastic to add now).


schwartzmx (approved):

schwartzmx:
Do we still need this @reshadi?

PR #60, opened 2023-04-06
Extend Channel to be pipeable

reshadi:
Sometimes we want to chain (pipe) channels together so that first all the event listeners of first channel recieve the events, and then next channel recieve it.

This is usually useful for either scheduling the second channel for later, or for transforming the events between channels.

This commit adds such capability to channels.

Also, separted Hook and Channel into their own bundle so that outside applications can use them independently.


packages/hook/test/Channel.test.ts
```diff
@@ -65,4 +65,65 @@ describe("test Channel", () => {
...
 
+  test("piped channels", (done) => {
+    const channel1 = new Channel<ChannelEvents>();
+    const channel2 = new Channel<ChannelEvents>();
+    const channel3 = new Channel<ChannelEvents>();
+    const fn1 = jest.fn<void, ChannelEvents['ev3']>();
+    const fn2 = jest.fn<void, ChannelEvents['ev3']>();
+
+    channel1
+      .pipe(channel2, task => {
```
schwartzmx, line 76: I could see this being nice for centralizing al_ui_event filtering and piping to specific channels which simplify subscribers having to check if `event = click` or other cases.  `UIChannel -> [ClickChannel, KeyChannel ...]`.   Could also see it's uses for other just general event driven narrowing and chaining

schwartzmx (approved):

packages/hook/test/Channel.test.ts
reshadi: Yes, for example everywhere that we are doing `requestIdleCallback` can be replaced by simply creating an `anync channel` that its events are fired only during idle time. That way, subscribers simply choose which phase they want to subscribe. 

In other words, this is a generic way of adding phases to channels and have certain event handlers run in a certain phase.

PR #62, opened 2023-04-07
Expose the network request url as a string

vsylee:
Proposing to simplify the type for the network request url and only expose it as a string. Current downstream usages convert to a string first.


packages/hyperion-autologging/src/ALNetworkPublisher.ts
```diff
@@ -26,7 +26,7 @@ type ALNetworkEvent = ALTimedEvent & Readonly<{
  */
 type RequestInfo = {
   method: string;
-  url: string | URL; // TODO: should we always send on of the types?
+  url: string;
```
reshadi, line 29: since usually parsing a url is tricky, how about we add an optional `parsedUrl` field and keep the `URL` value in that one as well?

reshadi (approved):

PR #64, opened 2023-04-19
A new algorithm to track flowlets in react

reshadi:
Two sets of changes are included in this commit:
1- a set of common hooks that have a function callback are intercepted in react. This set is then used to ensure those callbacks carry the flowlet of the render function forward.
2- In each render function, instead of relying on the prop extensions to detect the current flowlet, we use the the SurfaceContext to use the flowlet of the current surface. This flowlet is then picked up by other lifecycle methods or hooks and propagated.

I also updated the test app to ensure the full flowlet initialization is applied for all async browser api.

Together, this new algorithm give much bigger coverage for flowlets.

### DIFF
--- a/packages/hyperion-autologging/src/ALIReactFlowlet.ts
+++ b/packages/hyperion-autologging/src/ALIReactFlowlet.ts
@@ -1,0 +1,142 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+ */
+
+import TestAndSet from "@hyperion/hyperion-util/src/TestAndSet";
+
+import type * as Types from "@hyperion/hyperion-util/src/Types";
+
+import { assert } from "@hyperion/global";
+import * as IReact from "@hyperion/hyperion-react/src/IReact";
+import * as IReactComponent from "@hyperion/hyperion-react/src/IReactComponent";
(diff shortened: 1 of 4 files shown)


packages/hyperion-autologging/src/ALIReactFlowlet.ts
```diff
@@ -0,0 +1,142 @@
...
+          const ctx = useALSurfaceContext();
+          activeFlowlet = ctx.flowlet;
+          if (activeFlowlet) {
+            flowletManager.push(activeFlowlet)
+          }
+        });
+        fi.onValueObserverAdd(() => {
+          if (activeFlowlet) {
+            if (__DEV__) {
+              assert(
```
vsylee, line 130: Is there a way to get these to log to `FBLogger` in WWW? It would be good to track when these happen and see if we can plug the gaps

vsylee (approved):

packages/hyperion-autologging/src/ALIReactFlowlet.ts
reshadi: Yes, I had devised the plan for doing this later, but didn't get a chance to expose it. 
Basically, we should be able to replace the default logger here: 

https://github.com/facebook/hyperion/blob/main/packages/global/src/assert.ts#L14-L17

PR #66, opened 2023-04-25
Add alflowlet to the normal flowlets & network events for better

reshadi:
tracking

As we further harden the flowlet tracking, we might lose the link between a user interaction event and other events such as network events.
This commit adds a new internal ALFlowletManager that we separately track the flowlets created by AL itself. Then we attach them to the normal flowlets of each process to signal to them what "overlapped" with their flow.
In this way, we deterministically see the link between events and ui events.


packages/hyperion-autologging/src/ALType.ts
```diff
@@ -9,6 +9,7 @@ import { ALFlowlet, ALFlowletManager } from './ALFlowletManager';
 
 export type ALFlowletEvent = Readonly<{
   flowlet: ALFlowlet;
+  alflowlet?: ALFlowlet;
```
vsylee, line 12: nit: for consistency 

```suggestion
  alFlowlet?: ALFlowlet;
```

vsylee (approved):

PR #67, opened 2023-04-27
Added ClientSessionID that is tied to a session.

reshadi:
Sometimes applications need a single session id that links all pages of the same origin together. This id must survive across page loads.

Using browser's `sessionStorage`, this commit creates one such ID and apps can use in their logging safely across pages.


beltonzhong1 (approved):
LGTM

PR #68, opened 2023-05-09
Intercept useState and pass around flowlets to surface mutation events.

reshadi:
Sometimes a late async function (e.g. network response processing) may call the `setter` function returned by a `useState`. That action may then trigger more re-render, etc. resulting more mount/unmount events.

In these cases, we want the `alflowlet` to be assigned to the surface rendering flows so we can link user event to the (un)mount of the ui elements.

by intercepting and wrapping the setter function returned from `useState` we basically trigger the mechanism that marks the flowlets with known `alflowlet`.

I updated the test app to clearly see the link between async network load and the surface mutation event.


beltonzhong1 (approved):
LGTM

PR #69, opened 2023-05-09
Intercept Element.style and corresponding property methods.

reshadi:
This commit first intercepts the `Element.style` attribute, then adds interception for CSSStyleDeclaration and a few of its methods.

Also added a sample code as test to show how to block setting certain values.


beltonzhong1 (approved):
LGTM

PR #70, opened 2023-05-19
Ensure only plane functions are used for Function Components

reshadi:
There was a case in Meta code that a "special typeof" class component without a 'render' function was created and passed around. T153020999 (S339199) The existing code does not detect that and will assume it is a Function Component. As a result, we might incorrectly call hooks for such components, which causes React to throw.

The change in this commit will ensure that only "plain functions" are considered for functional component interception.

### DIFF
--- a/packages/hyperion-react/src/IReactComponent.ts
+++ b/packages/hyperion-react/src/IReactComponent.ts
@@ -185,13 +185,18 @@ export function init(options: InitOptions): void {
       /**
        * This is a react component, and can be a class component constructor
        * or functional component.
-       * In this case, we add the flowlet to the props and then
-       * when various lifecycle functions are called we can read the flowlet.
        *
        * Note that react itself has no types, and flow compiler has some built-in
        * hard coded types just to be able to handle react types.
        * (see all the React$* types in react.js)
        * So, sadly the code bellow effectively cannot rely on types much.
+       * 
+       * We need to check for two conditions:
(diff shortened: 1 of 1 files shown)


beltonzhong1 (approved):
LGTM

packages/hyperion-react/src/IReactComponent.ts
```diff
@@ -185,21 +185,27 @@ export function init(options: InitOptions): void {
...
        */
 
       // $FlowIgnore[prop-missing]
       const classComponentParentClass = component.prototype;
 
       if (
         classComponentParentClass instanceof ReactModule.Component ||
-        typeof classComponentParentClass?.render === 'function' // possibly created via React.createClass
+        typeof classComponentParentClass?.render === 'function' || // possibly created via React.createClass
+        Object.getPrototypeOf(Object.getPrototypeOf(classComponentParentClass)) // not a plane function, may be with some other lifecycle methods.
```
vsylee, line 208:
```suggestion
        Object.getPrototypeOf(Object.getPrototypeOf(classComponentParentClass)) // not a plain function, may be with some other lifecycle methods. 
```

vsylee (commented):
Can we also add an error guard so that future potential failures with React component interception does not crash the app using it? `AutoLogging.ts` could potentially take another option (e.g. `logger`, that we could pass `FBLogger` to from WWW) so that we still log that something went wrong.

PR #73, opened 2023-05-31
read ALSurfaceContext once per component

reshadi:
In some cases, calling the `useALSurface` in the render of every component may lead to infinit loops inside of react engine.

This change applies two fixes:
1- Only read the surface context once per component instance. It stores the value in the `this` of class components, or a ref (using `useRef`) in functional components
2- It avoids calling the `useContext` directly, and only optimistically read the `Context._value` internal value.

### DIFF
--- a/packages/hyperion-autologging/package.json
+++ b/packages/hyperion-autologging/package.json
@@ -25,8 +25,8 @@
     "@hyperion/hyperion-react": "*"
   },
   "peerDependencies": {
-    "@types/react": "^18.0.27",
-    "@types/react-dom": "^18.0.10"
+    "@types/react": "^18.2.7",
+    "@types/react-dom": "^18.2.4"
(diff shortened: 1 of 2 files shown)


vsylee (commented):
I ran into this error when trying to load T152801865- is it related?
image

vsylee:
> > I ran into this error when trying to load T152801865- is it related? image
> 
> Was this before the change in this PR? we don't use `useRef` somewhere else. The problem must be related to the previous issue we saw because the strange behavior of `ReactPropOverMenu`; it is a class component that does not inherit from the `Component` class, and also this class hierarchy does not have a `render` method! #70 was a fix for that.

Ah yeah seems I had incomplete changes, the latest changes look good!

vsylee (approved):

PR #74, opened 2023-06-07
Rename onArgs* and onValue* to onBeforeCallArgs* and onAfterReturnValue*

reshadi:
I got several feedback that the original names are not very descriptive and would prefer something that more explicitly indicate the callbacks will before function call or after return value is ready. Hence the rename here.
This is a breaing change, but since we still don't have an official versioning and releasing process, we'd better make this change now!

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -85,7 +85,7 @@ function ignoreInteractiveElement(node: HTMLElement) {
 const TrackedEvents = new Set<string>();
 
 let installHandlers = () => {
-  IEventTarget.addEventListener.onBeforeCallArgsObserverAdd(function (
+  IEventTarget.addEventListener.onBeforeCallObserverAdd(function (
     this: EventTarget,
     event,
     _listener,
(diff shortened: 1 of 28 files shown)


packages/hyperion-autologging/src/ALIReactFlowlet.ts
```diff
@@ -41,12 +41,12 @@ export function init(options: InitOptions) {
...
     IReactModule.useReducer
   ].forEach(fi => {
-    fi.onArgsMapperAdd(args => {
+    fi.onBeforeCallArgsMapperAdd(args => {
       args[0] = flowletManager.wrap(args[0], fi.name);
       return args;
     });
   });
-  IReactModule.useState.onValueMapperAdd(value => {
+  IReactModule.useState.onAfterReturnValueMapperAdd(value => {
```
vsylee, line 49: A few thoughts on the naming pattern:
- Suggest changing the order from `...Add` to `add...`, to follow similar patterns as `addEventListener`
- Suggest using the same `...BeforeX...`/`...AfterX...` phrase 

So maybe renaming to something like `addArgsMapperBeforeCall` and `addValueMapperAfterCall`?

vsylee (approved):
This looks fine, only the names are really long now. Thoughts on something like this?

- `onArgs(Observer|Mapper)Add` -> `onBeforeCall(Observer|Mapper)Add`
- `onValue(Observer|Mapper)Add` -> `onAfterCall(Observer|Mapper)Add`
- `onArgsAndValueMapperAdd` -> `onBeforeAndAfterCallMapperAdd`

PR #76, opened 2023-06-08
Link mount/unmount and request/response events directly

reshadi:
Linked request/response and mount/unmount events with their eventIndex.


reshadi:
> lgtm, is the idea to link these in the logging? This just sets the stage for accessing the prior linked event?

yes exactly. want to log the eventIndex of the corresponding pair directly.

PR #78, opened 2023-06-08
Fix incorrect alFlowlet assignments

vsylee:
Defaulting the assignment of `alFlowlet` to the `alFlowlet` corresponding to the top of the regular flowlet stack, when there aren't any active UI event flowlets, seems to be contributing to the problem of incorrect `alFlowlet`s getting set.


reshadi:
> Defaulting the assignment of `alFlowlet` to the `alFlowlet` corresponding to the top of the regular flowlet stack, when there aren't any active UI event flowlets, seems to be contributing to the problem of incorrect `alFlowlet`s getting set.

Did you test it with the am_al_react_flowlet gk on? This assumes that the best correct flowlet was already on the stack.

reshadi (approved):

PR #85, opened 2023-06-27
Fixed interception of Promises for Network events.

reshadi:
The al_network_* events assume Promise interception is already happened. We did have a empty call to `IPromise` to ensure that but it was removed during build time.

If we were calling `initFlowletTrackers` it would trigger Promise interception as a side effect. So, in some cases we were lukcy.

I noticed in some applications that we have console error for handling fetch. It turned out a) initFlowletTrackers was not called by default as part of AutoLogging, and b) call to require IPromise was dropped during build of bundles.

The code here fixes these issues and ensures we can handle `fetch` calls with our network events.


beltonzhong1 (approved):
LGTM

vsylee (approved):

schwartzmx (approved):

PR #88, opened 2023-07-18
Add flowlet ids

vsylee:
Changes made:
- Added a simple id setter (on initialization) and getter for the base Flowlet class. This is a starting point for being able to pass around and identify flowlets without having to rely on the full name string.
- Added an `onFlowletInit` hook to be able to track when new flowlet instances/ids are created

### DIFF
--- a/packages/hyperion-flowlet/src/Flowlet.ts
+++ b/packages/hyperion-flowlet/src/Flowlet.ts
@@ -2,14 +2,23 @@
  * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
  */
 
+import { Hook } from "@hyperion/hook";
+
+export const onFlowletInit = new Hook<(flowlet: Flowlet) => void>();
+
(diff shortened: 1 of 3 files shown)


packages/hyperion-flowlet/src/Flowlet.ts
```diff
@@ -2,14 +2,23 @@
...
 
+import { Hook } from "@hyperion/hook";
+
+export const onFlowletInit = new Hook<(flowlet: Flowlet) => void>();
+
+let flowletID: number = 0;
+
 export class Flowlet<T extends {} = {}> {
   readonly data: T;
+  private _id: number = flowletID++;
```
reshadi, line 13: instead you could just say
```
readonly id: number = flowletID++;
```

That way, the value stays fixed and you don't need a getter to protect the value.

reshadi (approved):

reshadi (approved):

PR #89, opened 2023-07-26
Surface config channel should use publisher provided channel

schwartzmx:
- Since the channel provided is _currently_ only relevant when publishing mutations,  utilize the one provided when configuring the publisher
	- This way there's no way to misconfigure the channels when different ones are provided, since there's an interdependency between the two to enable the publisher

Validated both when publisher config is provided and not provided that things seem to function as expected.

Mutation events:
image

Validated with publisher config omitted surfaces are still present and no mutation events are emitted.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -73,7 +73,8 @@ export type InitOptions = Types.Options<
     IReactModule: IReact.IReactModuleExports;
     IJsxRuntimeModule: IReact.IJsxRuntimeModuleExports;
     domFlowletAttributeName?: string;
-    channel: ALChannel;
+    // Only used if ALSurfaceMutationPublisher is enabled as well.
+    channel?: ALChannel;
     disableReactDomPropsExtension?: boolean;
   }
 >;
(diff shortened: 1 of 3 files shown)


packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -71,7 +71,8 @@ export type InitOptions = Types.Options<
     IReactModule: IReact.IReactModuleExports;
     IJsxRuntimeModule: IReact.IJsxRuntimeModuleExports;
     domFlowletAttributeName?: string;
-    channel: ALChannel,
+    // Only used if ALSurfaceMutationPublisher is enabled as well.
+    channel?: ALChannel,
```
reshadi, line 75: Just note in the future, we might have to move the definition of `ALChannel` out of this module, but at the same time, it feels odd to have `ALSurface` depend on `ALSurfaceMutationPublisher`.

reshadi (approved):

packages/hyperion-autologging/src/ALSurface.ts
schwartzmx: I'm definitely open to doing this a different way.   We could have the publisher enabled within the surface config itself,  like a sub init option,  since they are closely related.   Without surfaces the publisher doesn't make much sense either.

reshadi: I think the way it is currently is ok and you can merge it. I can think that we might have applications that are not react based and they might directly add DOM attributes for the surfaces. In those case, we still want to have publisher work as it is now, but just that we might have different way of handing the surface itself. I don't yet have a strong conviction which way to go. Just wanted to share with you the alternatives to consider. 
But as of now, let's land this and fix the internal issue we have sooner. We can revisit this easily later since this is just a type.

reshadi (approved):

PR #90, opened 2023-07-26
Generate al_flowlet_event

reshadi:
This event will allow us to store flowlet and other related information for them in the log stream. It also allow visualizer to track flowlet creation and visualize them.


packages/hyperion-react-testapp/src/AutoLoggingWrapper.ts
```diff
@@ -53,6 +53,7 @@ export function init() {
     'al_network_request',
     'al_network_response',
     'al_network_response',
+    'al_flowlet_event',
```
schwartzmx, line 56: Can you see these emitted in the test app?  I checked out the branch but not seeing any.

reshadi: yes, I do see them. Did you do a full build after checking out? 

image

schwartzmx (approved):

PR #91, opened 2023-07-27
add label support

beltonzhong1:
Checking the aria-label attribute if innerText is empty. The label is a reliable description of what the element does included for accessibility.

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -192,6 +192,10 @@ export function getElementName(
     if (name != null) {
       return name;
     }
+    const label = nextElement.getAttribute('aria-label');
+    if (label != null) {
+        return label;
+    }
     nextElement = nextElement.parentElement;
   }
   return null;
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -187,6 +187,24 @@ export type ALElementNameResult = Readonly<{
   source: 'innerText' | 'aria-label' | 'aria-labelledby'| 'aria-description' | 'aria-describedby';
 }>;
 
+// Takes a space-delimited list of DOM element IDs and returns the inner text of those elements joined by spaces
+function getTextFromElementsByIds(ids: string): string | null {
+  const fullText = ids.split(' ').map(id => {
+    const element = document.getElementById(id);
+    if (element != null) {
+      const text = extractInnerText(element);
+      if (text != null) {
+          return text;
+      }
+    }
+    return null;
+  }).join(' ');
+  if (fullText != null && fullText !== '') {
+    return fullText;
+  }
+  return null;
+}
+
 export function getElementName(
   element: HTMLElement,
   // Whether to try to resolve if element name is a result of Fbt translation
@@ -196,25 +214,30 @@ export function getElementName(
 
     const labelledBy = nextElement.getAttribute('aria-labelledby');
     if (labelledBy != null) {
-      const fullLabel = labelledBy.split(' ').map(labelledBy => {
-        const labelElement = document.getElementById(labelledBy);
-        if (labelElement != null) {
-          const labelText = extractInnerText(labelElement);
-          if (labelText != null) {
-            return {
-              text: labelText,
-            };
-          }
-        }
-        return null;
-      });
-      const fullLabelText = fullLabel.map(label => label?.text).join(' ');
-      if (fullLabelText != null && fullLabelText !== '') {
+      const labelText = getTextFromElementsByIds(labelledBy);
+      if(labelText != null && labelText !== '') {
         return {
-          text: fullLabelText,
-          source: 'aria-labelledby',
+          text: labelText,
+          source: 'aria-labelledby'
         };
       }
+      // const fullLabel = labelledBy.split(' ').map(labelledBy => {
+      //   const labelElement = document.getElementById(labelledBy);
+      //   if (labelElement != null) {
+      //     const labelText = extractInnerText(labelElement);
+      //     if (labelText != null) {
+      //         return labelText
+      //     }
+      //   }
+      //   return null;
+      // });
+      // const fullLabelText = fullLabel.join(' ');
+      // if (fullLabelText != null && fullLabelText !== '') {
+      //   return {
+      //     text: fullLabelText,
+      //     source: 'aria-labelledby',
+      //   };
+      // }
     }
 
     const label = nextElement.getAttribute('aria-label');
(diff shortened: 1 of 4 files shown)


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -192,6 +192,10 @@ export function getElementName(
     if (name != null) {
       return name;
     }
+    const label = nextElement.getAttribute('aria-label');
```
reshadi, line 195: 3 questions:
1- what about the `aria-labledby` support? We should detect that and find the corresponding element. 
2- should we give this more priority than the innerText? 
3- how do we communicate to outside of this function that this value is definitely not a UGC?

schwartzmx: I think the source would be useful for debugging as well so we know if it's `innerText`, `aria-label` or other source of the text,  potentially stuff this in metadata.

Maybe this function can return a result object `{text: string, source: 'innerText' | 'aria-label' | ...}`

packages/hyperion-autologging/src/ALType.ts
```diff
@@ -43,4 +43,9 @@ export type ALReactElementEvent = Readonly<{
...
   flowletManager: ALFlowletManager;
   domSurfaceAttributeName: string;
-}>;
\ No newline at end of file
+}>;
+
+export type ALElementNameResult = Readonly<{
+  text: string,
+  source: 'innerText' | 'aria-label' | 'aria-labelledby';
+}>;
```
reshadi, line 51: Move this type to packages/hyperion-autologging/src/ALInteractableDOMElement.ts. No one else needs it.
In general, it is better to keep the types as close to the place they are produced and avoid one giant file with all the types. Imagine if one day we remove packages/hyperion-autologging/src/ALInteractableDOMElement.ts, and replace with something new, then we want to have the least amount of places to clean up. That's usually a good test of where to put types. 
packages/hyperion-autologging/src/ALType.ts is good for types that are global and needed in many places.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -29,6 +29,8 @@ type ALUIEvent<T = EventHandlerMap> = {
     element: HTMLElement | null,
     // Element text extracted from element
     elementName?: string | null,
+    // The source attribute where we got the elementName from
+    elementNameSource?: string | null,
```
reshadi, line 33: I thought we want this data to be part of the metadata, but if you want to add it as a explicit field, better be more precise
`elementNameSource: ALElementNameResult['source'] | null;`

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi, line 1: also add `true` to line 154 to start from interactive elements.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -185,13 +186,29 @@ const extractInnerText = (element: HTMLElement | null): string | null => {
...
-): string | null {
+): ALElementNameResult | null {
   let nextElement: HTMLElement | null = element;
   while (nextElement && nextElement.nodeType === Node.ELEMENT_NODE) {
+    const label = nextElement.getAttribute('aria-label');
+    if (label != null && label !== '') {
+
+        return {text: label, source: 'aria-label'};
+    }
+    const labelledBy = nextElement.getAttribute('aria-labelledby');
```
reshadi, line 197: I noticed in the spec that `aria-labeledby` takes precedence over other methods, so you should check this first before `aria-label` to be more accurate: 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#:~:text=aria%2Dlabelledby%20takes%20precedence%20over%20all%20other%20methods

Furthermore, according to spec, there could be multiple values separated by space that should be picked up and combined. 

There is an alternative as well `aria-descript` and `aria-describedby` which has similar meaning.: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -182,16 +182,83 @@ const extractInnerText = (element: HTMLElement | null): string | null => {
...
+    if (labelledBy != null) {
+      const fullLabel = labelledBy.split(' ').map(labelledBy => {
+        const labelElement = document.getElementById(labelledBy);
+        if (labelElement != null) {
+          const labelText = extractInnerText(labelElement);
+          if (labelText != null) {
+            return {
+              text: labelText,
+            };
```
reshadi, line 206: you don't need to create yet another object here. Just return `labelText`

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -182,16 +182,83 @@ const extractInnerText = (element: HTMLElement | null): string | null => {
...
+          }
+        }
+        return null;
+      });
+      const fullLabelText = fullLabel.map(label => label?.text).join(' ');
```
reshadi, line 211: I think you also need to filter null, or better yet, just return '' (empty string) instead of null so that when you call `.join` it works well.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -182,16 +182,83 @@ const extractInnerText = (element: HTMLElement | null): string | null => {
...
+          const descText = extractInnerText(descElement);
+          if (descText != null) {
+            return {
+              text: descText,
+            };
+          }
+        }
+        return null;
+      });
+      const fullDescText = fullDesc.map(desc => desc?.text).join(' ');
```
reshadi, line 247: same comment for here. In fact, you can turn this into a function that receives the whole string, with spaces, and returns the final string.

packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -94,7 +94,10 @@ export function publish(options: InitOptions): void {
           if (cacheElementReactInfo) {
             const elementInfo = ALElementInfo.getOrCreate(node);
             reactComponentData = elementInfo.getReactComponentData();
-            elementName = getElementName(node);
+            const elementNameResult = getElementName(node);
+            if(elementNameResult) {
+              elementName = elementNameResult.text;
+            }
```
reshadi, line 100: we would probably want the source here as well. 
Consider that when we receive them in WWW, we will will need to look at source + FBT and other infor and decide to keep the value or drop/encode it.

reshadi (commented):
Very close! Thanks for adding the extra pieces. Just a few changes and we are ready to go

reshadi (approved):

PR #98, opened 2023-08-02
Enhancedelementtext

reshadi:
Cleanup the types and unify handing of the element text.
Exposed an optional config to allow further control of element text externally.


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -27,43 +27,34 @@ type ALUIEvent<T = EventHandlerMap> = {
...
-  ALReactElementEvent &
-  {
-    autoLoggingID: ALID | null,
-    surface?: string | null,
-  }
+  Omit<ALUIEventCaptureData, 'captureTimestamp'> &
+  ALTimedEvent
 >;
 
 export type ALUIEventData = Readonly<
```
schwartzmx, line 60: In this file we use `getInteractable` with `true` in the capture phase but not in the bubble phase,  maybe that can cause some weirdness in the output

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -182,113 +174,127 @@ const extractInnerText = (element: HTMLElement | null): string | null => {
...
+  for (
+    let nextElement: HTMLElement | null = element;
+    nextElement && nextElement.nodeType === Node.ELEMENT_NODE;
+    nextElement = nextElement.parentElement
+  ) {
 
-    const name = extractInnerText(nextElement);
-    if (name != null && name !== '') {
-      return {text: name, source: 'innerText'};
+    const text = getTextFromElementsByIds(domSource, 'aria-labelledby')
```
schwartzmx, line 274: `nextElement` is not passed to these extraction methods,  looks like we just use the same `directSource: element` each time.

Not sure if we want to add each iteration's (after the first) element as `indirectSources` or not.

schwartzmx: It might actually be better to not loop, but open to thoughts.   In some cases we may walk way up the dom and get some text that's totally unrelated

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi: I will send an update soon to make sure the logic is shared between capture and buble, and test to see if data is correct. Thanks for noticing the difference.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
reshadi: Oh shoot! This was missing! I should update the `domSource.directSource = nextElement` in each iteration. 

the looping here is necessary because of the priority of different attributes.

schwartzmx (approved):
Changes look good,  thanks for the detailed comments as well

PR #99, opened 2023-08-04
Added testAndSet to FunctionInterceptor for conenience.

reshadi:
In many cases, we check the data value once and then set it. This helper method makes it more concise.

Also removed __DEV__ for function component name.

### DIFF
--- a/packages/hyperion-core/src/FunctionInterceptor.ts
+++ b/packages/hyperion-core/src/FunctionInterceptor.ts
@@ -427,6 +427,13 @@ export class FunctionInterceptor<
     }
     this.data[dataPropName] = value;
   }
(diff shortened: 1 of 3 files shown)


schwartzmx (approved):
We also have https://github.com/facebook/hyperion/blob/main/packages/hyperion-util/src/TestAndSet.ts - is it possible to extend to handle this use case as well?

reshadi:
> We also have https://github.com/facebook/hyperion/blob/main/packages/hyperion-util/src/TestAndSet.ts - is it possible to extend to handle this use case as well?

What do you mean to extend to handle this case? This commit is adding the value to the FunctionInterceptor, that standalone class for standalone values.

PR #102, opened 2023-08-17
Rename alFlowlet to uiEventFlowlet

vsylee:
Changes made:
- Renamed `alFlowlet` to `uiEventFlowlet`, since these are flowlets that map directly to UI events and are used to link generic AL events with the triggering UI event
- Removed `alFlowlet` from channel events since it can be read directly from `flowlet`


reshadi (approved):
Please double check that removing the uiEventFlowlet from the event will not break the logic. 
I think this is the right thing to do, however, if between the time event is fired, and the time it is actually logged, the value of uiEventFlowlet changes on the corresponding flowlet, we will lose the correct value.

PR #103, opened 2023-08-18
Use targetElement for resolving React info

schwartzmx:
- With the change to use interactable element as source for better text extraction we seem to now run into cases where the interactable element does not have fiber information and may be installed by react itself, and may only contain the prop `_reactListening*` and no `_reactFiber` available.
- This change just keeps a ref to targetElement and uses it for resolving react information, leaving the other uses of the resolved interactableElement as is.


reshadi (approved):
The change is good. I wonder since `event.target` is well known, was it better to name the new field `targetElement` instead of `sourceElement`

schwartzmx:
`sourceElement` -> `targetElement` in last push

reshadi (approved):

PR #104, opened 2023-08-19
Generalize detection of attribute event handlers

reshadi:
The code was relying on a hardcoded `eventNamesMap` to detect `on$[eventname}` from `eventName`.
The code was already missing new events such as key events.

This fix removes the hardcoded table and instead uses a type safe approach to directly detect existance of `on*` event handlers.

Also added some simple tests so that going forward we can test each such change.


schwartzmx (approved):
nice this is much better, then we just set these attributes for all tracked events as well

PR #105, opened 2023-08-19
Add default .metadata to all events

reshadi:
Often application code might want to extend the standard AL events and add data to be logged.

Although a recomended way would be to extend the event types themselves, but in some simple cases, it is easier to just add the new data to a default `.metadata` object that is later logged along with the rest of data.

This commit adds this feature, along with some tests. Specially for ui events, the metadata of the capture and bubble events is eventually merged into one and passed to the actual `al_ui_event`.

Later we will properly add supporting metadata for surfaces as well.

### DIFF
--- a/packages/hyperion-autologging/src/ALElementInfo.ts
+++ b/packages/hyperion-autologging/src/ALElementInfo.ts
@@ -6,7 +6,8 @@
 
 import { getReactComponentData_THIS_CAN_BREAK } from './ALReactUtils';
 import type { ReactComponentData } from './ALReactUtils';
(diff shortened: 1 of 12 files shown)


beltonzhong1 (approved):
LGTM

reshadi:
> Looks good, I know we typically use `string` values for metadata, but wondering if it makes sense to relax that some, or support union of other primitives

Initially I thought using a generic type for this, similar to Flowlet data, but then I realized it makes a lot of things complicated. Using the union of types is possible, but makes reading stuff back from it rather inconvenient. I assumed in all cases, we eventually convert them to string for storage any ways, right? So, we'd better be explicit from start so that call sites can chose their own correct stringify algorithm when needed.

PR #106, opened 2023-08-21
Changes testing enablement with ssr/etc.

schwartzmx:


packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -67,7 +67,9 @@ export function init(options: InitOptions): boolean {
     setComponentNameValidator(options.componentNameValidator);
   }
 
-  initFlowletTrackers(options.flowletManager);
+  if (document != null && document.createElement != null) {
```
reshadi, line 70: to be safe, import `globalScope` and use that one. If a global variable is not defined in the scope, it can cause an exception in strict mode. 

```
import globalScope from "@hyperion/global/src/global";

...
if (global?.document?.createElement != null) {
...
}
```

reshadi (approved):

packages/hyperion-autologging/src/AutoLogging.ts
schwartzmx: Tried that at first but couldn't access `document`,
```
Property 'document' does not exist on type 'typeof globalThis | (Object & OurGlobal & NodeGlobal) | Window'.
  Property 'document' does not exist on type 'Object & OurGlobal & NodeGlobal'.ts(2339)
  ```
Typed as
`const global: typeof globalThis | (Object & OurGlobal & NodeGlobal) | Window`
  is there a good way to make it possible to access document if available?

schwartzmx: @reshadi should I apply the above check?  What do you think?

reshadi: yes, that is good to apply. please go ahead to do.

PR #107, opened 2023-08-22
Change interactable algorithm to detect atag, and also check for assi…

schwartzmx:
…gned handlers

- Changes the algorithm to walk up the tree itself, rather than rely on `closest`
- When walking up the tree, check if the element meets our selector or if there is a handler installed/assigned to the element
- Added tests for various scenarios, but there may be more we should test


reshadi (approved):
Nice job with the test!

PR #108, opened 2023-08-23
handle on* event handlers

reshadi:
Enabled handling of `element.on* = (...)=>{ ... }`  both for wrapping the callback and passing the right flowlet, and for detecting interactable elements.


src/index.ts
```diff
@@ -22,6 +22,7 @@ export * as IEventTarget from "@hyperion/hyperion-dom/src/IEventTarget";
 export * as IElement from "@hyperion/hyperion-dom/src/IElement";
 export * as IHTMLElement from "@hyperion/hyperion-dom/src/IHTMLElement";
 export * as ICSSStyleDeclaration from "@hyperion/hyperion-dom/src/ICSSStyleDeclaration";
+// export * as IGlobalEventHandlers from "@hyperion/hyperion-dom/src/IGlobalEventHandlers";
```
reshadi, line 25: I left this here as a reminder and for consistency with other bundler, but as of now we don't need to worry about exposing this to the outside world!

packages/hyperion-dom/src/EventHandlerAttributeInterceptor.ts
```diff
@@ -2,23 +2,54 @@
...
+}
+
+/**
+ * Use this function when the event handler can be set via node.onX
+ * as well as node.setAttribute("X", ....) or node.getAttributeNode("X").value = ...
+ * This mostly applies to Elements. 
+ * In these cases, one need to cover both raw (text) and processed (function) values
+ * for the handler. 
+ */
+export function interceptElementEventHandlerAttribute<
```
reshadi, line 46: This is the more advanced case where someone could add `setAttribute("onclick", ....) to an element, but it very rare and would need to further test the implementation. So for now, keep the old simple one.

packages/hyperion-flowlet/src/Index.ts
```diff
@@ -20,115 +20,115 @@ export function initFlowletTrackers(flowletManager: FlowletManager) {
...
-    // IGlobalEventHandlers.onoffline,
-    // IGlobalEventHandlers.ononline,
-    // IGlobalEventHandlers.onpagehide,
-    // IGlobalEventHandlers.onpageshow,
-    // IGlobalEventHandlers.onpopstate,
-    // IGlobalEventHandlers.onrejectionhandled,
-    // IGlobalEventHandlers.onstorage,
-    // IGlobalEventHandlers.onunhandledrejection,
-    // IGlobalEventHandlers.onunload,
+    IWindow.ondevicemotion,
```
reshadi, line 23: I realized we are also not handling these cases for flowlet wrapping.

packages/hyperion-autologging/test/ALInteractableDOMElement.test.ts
```diff
@@ -108,4 +108,28 @@ describe("Text various element text options", () => {
...
+      <div id="addEventListener"></div>
+    `);
+
+    ALInteractableDOMElement.trackInteractable("click");
+
+    let node = document.getElementById("attribute");
+    expect(node).not.toBeNull();
+    if (node) {
+      node.onclick = () => { };
+      expect(node.getAttribute("data-clickable")).toBe("1");
```
schwartzmx, line 124: minor but could also include a test setting `node.onclick = null` and see if the attribute gets removed.

packages/hyperion-flowlet/src/Index.ts
schwartzmx: nice find

schwartzmx (approved):
lgtm,  but there's a conflict in the test from other PR being merged

reshadi:
> 

Yes, will resolve and push again.

PR #109, opened 2023-08-25
Ensure react interception is enabled for tracking interactable elements

reshadi:
In order to track custom events on DOM elements, we should make sure that `IReactComponent.init()` is already called.
This commit fixes that.
In the past, if `disableReactDomPropsExtension` was false, the `ALSurface` would have initialized `IReactComponent` by itself. But the code was not resilient enough to work independently.


reshadi:
> Possibly the `installHandlers` tries to get called before init is called in the text config?

No the problem was that I had added react init options to ALInteractableDOMElement.init, but had not updated the test code to pass it.
Instead, I separated the react options into a field for all functions and have it passed to `AutoLogging.init` function. That we can we can call react interception directly at the top level

schwartzmx (approved):

PR #113, opened 2023-08-29
Onargsandvalueobserver

reshadi:
Introduced a more reliable interface to the `FunctionInterceptor` and then use it for better interception of react api.
See each commit's message for more details.


packages/hyperion-react/src/IReactFlowlet.ts
```diff
@@ -111,31 +111,49 @@ export function init<
...
-            flowletManager.pop(activeFlowlet);
+        // let activeFlowlet: FlowletType | undefined | null;
+        // fi.onArgsObserverAdd(props => {
+        //   activeFlowlet = flowletPusher(props);
+        // });
+        // fi.onValueObserverAdd(() => {
+        //   if (activeFlowlet) {
+        //     flowletManager.pop(activeFlowlet);
+        //   }
+        // });
```
vsylee, line 147: Remove

packages/hyperion-react/src/IReactFlowlet.ts
```diff
@@ -111,31 +111,49 @@ export function init<
...
-          flowletManager.pop(activeFlowlet);
+      // let activeFlowlet: FlowletType | undefined | null;
+      // method.onArgsObserverAdd(function (this: ComponentWithFlowlet) {
+      //   activeFlowlet = flowletPusher(this.props);
+      // });
+      // method.onValueObserverAdd(function (this: ComponentWithFlowlet) {
+      //   if (activeFlowlet) {
+      //     flowletManager.pop(activeFlowlet);
+      //   }
+      // });
```
vsylee, line 122: Remove

vsylee (approved):
Thanks for implementing this! I will update the Graph API case in WWW to use this new API as well

PR #115, opened 2023-08-30
Added ALCustomEvent

reshadi:
Enable users to send custom events that can be logged. Also fixed a missing field in the metadata of the UI events indicating if bubble did timeout.

### DIFF
--- a/packages/hyperion-autologging/src/ALCustomEvent.ts
+++ b/packages/hyperion-autologging/src/ALCustomEvent.ts
@@ -7,6 +7,9 @@
 import type { Channel } from "@hyperion/hook/src/Channel";
 
 import { ALLoggableEvent, ALMetadataEvent, ALOptionalFlowletEvent, ALTimedEvent } from "./ALType";
+import performanceAbsoluteNow from "@hyperion/hyperion-util/src/performanceAbsoluteNow";
+import * as ALEventIndex from "./ALEventIndex";
+import { ALFlowletDataType, ALFlowletManager } from "./ALFlowletManager";
 
 export type ALCustomEventData =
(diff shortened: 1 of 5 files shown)


packages/hyperion-autologging/src/ALCustomEvent.ts
```diff
@@ -0,0 +1,23 @@
...
+'use strict';
+
+import type { Channel } from "@hyperion/hook/src/Channel";
+
+import { ALLoggableEvent, ALMetadataEvent, ALOptionalFlowletEvent, ALTimedEvent } from "./ALType";
+
+export type ALCustomEventData =
+  ALMetadataEvent &
+  ALOptionalFlowletEvent &
+  ALTimedEvent &
```
schwartzmx, line 14: I was just thinking for some of these more "core" fields like eventTimestamp and eventIndex etc.. do you think we should control adding them to the event ourselves?  Just so we keep these consistent across all events?

schwartzmx (approved):
overall looks good,  just one thought

packages/hyperion-autologging/src/ALCustomEvent.ts
reshadi: Excellent question! I have did separate the module and only added a single type to it now, but my thought was to add a function that will fill out the rest of the fields. I only was concerned about the event index part. I was thinking if people wanted to generate events that may not be loggable, it will increase the event index . 
But now I think we can add an extra functions argument to give control for that as well. I'll add the function and export it externally. 

Note that one can still choose to directly call `channel.emit`, but we can just give them a mechanism for simplicity.

PR #116, opened 2023-09-01
Visualizer

reshadi:
Initial setup and a sample component for the new visualizer package.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,6 +15,9 @@
     "**/node_modules": true,
     "**/*.js": {
       "when": "$(basename).ts"
+    },
+    "**/*.jsx": {
+      "when": "$(basename).tsx"
     }
   }
 }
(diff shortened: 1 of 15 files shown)


.vscode/settings.json
```diff
@@ -15,6 +15,9 @@
     "**/node_modules": true,
     "**/*.js": {
       "when": "$(basename).ts"
+    },
+    "**/*.jsx": {
```
schwartzmx, line 19: Ah this is maybe what I was having trouble with when trying to set up the package and import tsx component into test app.

schwartzmx (approved):
thanks for setting this up,  we can merge and I can port what I was working on into this

.vscode/settings.json
reshadi: I doubt that. This setting is to tell VS code don't show the generated JSX files if you find a file with the same basename but with TSX extension. 
I think your problem might have been that `jsx: react` option in the `tsconfig.json`

schwartzmx: Ah I see,  I had `"jsx": "react-jsx"`,  and thought I tried the other variations as well,  but maybe some other misconfiguration on my side

PR #117, opened 2023-09-01
Change global window check,  Window not available in server

schwartzmx:
`instanceof Window` can't be executed in server env as `Window` doesn't exist.

Tested with this and seems to work as expected.

### DIFF
--- a/packages/hyperion-autologging/src/AutoLogging.ts
+++ b/packages/hyperion-autologging/src/AutoLogging.ts
@@ -71,7 +71,7 @@ export function init(options: InitOptions): boolean {
     setComponentNameValidator(options.componentNameValidator);
   }
 
-  if (global instanceof Window && global?.document.createElement != null) {
+  if (typeof global !== 'undefined' && (global as Window)?.document?.createElement != null) {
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -71,7 +71,7 @@ export function init(options: InitOptions): boolean {
     setComponentNameValidator(options.componentNameValidator);
   }
 
-  if (global instanceof Window && global?.document.createElement != null) {
+  if (global && (global as Window)?.document?.createElement != null) {
```
reshadi, line 74: This is what I was talking about the other day as well.
To be fully safe, you should do it this way:
```
if (typeof global !== 'undefined' && (global as Window)?.document?.createElement != null)
```

reshadi (approved):

packages/hyperion-autologging/src/AutoLogging.ts
schwartzmx: Ah,  thanks.  Done in latest push.

PR #121, opened 2023-09-06
Fix __DEV__ value.

reshadi:
We need to ensure __DEV__ is always defined since a lot of the code directly uses it. We also should make sure this code runs ok in github as well as the envirinment into which the code is imported.

Fixed the assignment to only update it to false in the JEST environment.

### DIFF
--- a/packages/global/src/global.ts
+++ b/packages/global/src/global.ts
@@ -26,8 +26,8 @@ const globalScope =
           {};
 
 
-if (typeof __DEV__ !== "boolean") {
-  (globalThis as OurGlobal).__DEV__ = true;
+if (typeof __DEV__ === "undefined") {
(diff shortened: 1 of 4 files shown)


packages/global/src/global.ts
```diff
@@ -36,6 +36,7 @@ if (typeof __DEV__ !== "boolean") {
     )
   ) {
     global["__DEV__"] = true;
+    console.warn('SET DEV TO TRUE')
```
vsylee, line 39: Maybe helps make it more descriptive, in case we see it happening outside of Jest environments

```suggestion
    console.warn('SET DEV TO TRUE FOR JEST');
```

vsylee (approved):

packages/global/src/global.ts
reshadi: I noticed that makes test outputs very noisy. I was planning to remove it for the final code.

PR #122, opened 2023-09-06
Avoid background tasks until necessary.

reshadi:
If we are not running in reall application (usually browser env.) we may not need some background tasks done.

This commit fixes two things:

* `SessionPersistentData` will only scheduling saving to disk if the feature is available (in browser). Otherwise, avoid unnecessary work.
* `FlowletManager` will only schedule cleanup when the stack size get bigger than a certain threshold. In normal cases, the pairs of push/pops should avoid this situation.


schwartzmx (approved):
lgtm

PR #123, opened 2023-09-06
Very rough draft of prototyping with cytoscape and graphing AL events

schwartzmx:
-  Lots of TODOs,  and need to abstract how links are created between events
- Since theres potentially multiple edges, and a need to lookback,  we need to probably come up with a better way to do the linking,  the current way is pretty hacky to say the least

Will iterate on this further now that something is working and clean up a bunch of things,  but pushing for now as I'll be out of town for a bit.

https://github.com/facebook/hyperion/assets/6064408/c9c81b3f-1257-4ba5-a8ca-5c51861001c2


packages/hyperion-autologging-visualizer/src/component/ALSessionGraph.react.js
reshadi, line 1: I updated the .gitignore, so you should no longer see the .js and .jsx files for checking.

packages/hyperion-autologging-visualizer/src/component/cytoscape-elk.d.ts
reshadi, line 1: why are these needed? There were no type for them?

packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -151,3 +153,25 @@ export function getSurfaceRenderer(defaultALSurfaceHOC?: ALSurface.ALSurfaceHOC)
   );
   return renderer;
 }
+
+
+/**
+ * Gets the init options passed when initializing AutoLogging.
+ * Can be useful to get configured channels, registered events, and other information after framework initialization.
+ */
+export function getInitOptions(): InitOptions {
```
reshadi, line 162: It looks like you mainly need the `channel` from the init options. 
Another , perhaps more reliable approach is to do something similar to the following where the application explicitly passes the channel to the component. 
https://github.com/facebook/hyperion/blob/main/packages/hyperion-autologging-visualizer/src/component/ElementTextTooltip.react.tsx

Effectively, we just need to call the following from the target application that prepares the visualizer. 
https://github.com/facebook/hyperion/blob/main/packages/hyperion-autologging-visualizer/src/index.ts

reshadi (approved):
since there is no publicly visible change, I approve so we can call iterate on this. 
Thanks for setting it up.

reshadi:
So the build is breaking and it seems like we are hitting the following issue: Apparently CRA is dead and we are supposed to move off of it. 

https://github.com/facebook/create-react-app/issues/13080#issuecomment-1623774848

packages/hyperion-autologging-visualizer/src/component/cytoscape-elk.d.ts
schwartzmx: Yeah no type,  but still even with these `.d.ts` files compiler was still complaining about them being typed as `any`.

packages/hyperion-autologging/src/AutoLogging.ts
schwartzmx: It may also be useful to know which events are being emitted on the channel (e.g. UIPublisher registered events) so that the visualizer can itself set up listeners for those events,  but yes mostly just need channel.

schwartzmx:
@reshadi let me look into the CRA deprecation and how to migrate away

schwartzmx:
@reshadi started migrating cra to vite in this diff https://github.com/schwartzmx/hyperion/pull/1
everything seems to work as expected,  _except_ ALSurface wrapping is not setting attributes,  therefore no mutation events when running test app.  Still need to dig into why.

reshadi:
> @reshadi started migrating cra to vite in this diff [schwartzmx#1](https://github.com/schwartzmx/hyperion/pull/1) everything seems to work as expected,~_except_ ALSurface wrapping is not setting attributes, therefore no mutation events when running test app. Still need to dig into why.~ Fixed in latest push, will try to separate these changes if needed.
> 
> Also on your question earlier, looks like they support node server plugin, https://www.npmjs.com/package/vite-plugin-node

This is great! If you're ready to land, no need to then separate the migration to Vite if that's too much work. You can land and we pick it up from there.

schwartzmx:
@reshadi  have not had much time to further iterate on this,  should we hold off on merging until some things can be cleaned up or just roll forward?

PR #124, opened 2023-09-09
Support regions (hidden surfaces)

reshadi:
Added support for hidden surfaces (regions). Optimized publishing surface events. Fixed flowlet assignment to (un)mount events. 

Nested surfaces will only reference parent surfaces
Nested regions will reference both parent surfaces and parent regions. 

See individual commits for more details.


packages/hyperion-autologging/src/ALFlowletManager.ts
```diff
@@ -14,6 +14,7 @@ import { Flowlet } from "@hyperion/hyperion-flowlet/src/Flowlet";
  */
 export interface ALFlowletDataType {
   surface?: string;
+  region?: string;
```
vsylee, line 17: Would like to propose renaming these to map to the surface capabilities. If I understand correctly, each field should be a path of surfaces that filter for one of the surface capabilities. Then maybe something like `interactionSurface` and `mutationSurface`?

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -11,52 +11,74 @@ import * as IReact from "@hyperion/hyperion-react/src/IReact";
...
+  ALSurfaceCapability.TrackInteraction,
+  ALSurfaceCapability.TrackMutation,
+];
+const AllSurfaceCapabilityValuesString = AllSurfaceCapabilityValues.map(c => ALSurfaceCapability[c]).join(',');
+function surfaceCapabilityToString(capability?: ALSurfaceCapability): string {
+  if (!capability) {
+    return AllSurfaceCapabilityValuesString;
+  }
+
+  return AllSurfaceCapabilityValues.filter(c => (capability & c) != 0).map(c => ALSurfaceCapability[c]).join(',');
```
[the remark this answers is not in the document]
reshadi: yes agreed, will fix

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -11,52 +11,74 @@ import * as IReact from "@hyperion/hyperion-react/src/IReact";
...
+  TrackMutation = 1 << 1, // report mount/unmount of this surface
+  // TrackVisibility = 1 << 2, // track when mounted surface's visibility changes
+  // TrackBoundingRect = 1 << 3, // Report the size of the bounding rect of the surface on the screen.  
+}
+
+const AllSurfaceCapabilityValues = [
+  ALSurfaceCapability.TrackInteraction,
+  ALSurfaceCapability.TrackMutation,
+];
+const AllSurfaceCapabilityValuesString = AllSurfaceCapabilityValues.map(c => ALSurfaceCapability[c]).join(',');
```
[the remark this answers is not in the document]
reshadi: TS enums are cool!

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -180,46 +206,87 @@ function setupDomElementSurfaceAttribute(options: InitOptions): void {
...
+      domAttributeValue,
+    };
+
+    if (options.channel && !proxiedContext) {
       const { channel } = options;
       // Emit surface mutation events on mount/unmount
+      const metadata = props.metadata ?? {}; // Note that we want the same object to be shared between events to share the changes.
+      metadata.original_flowlet = flowlet.getFullName();
+
+      metadata.surface_capability = surfaceCapabilityToString(props.capability);
```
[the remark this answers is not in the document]
reshadi: This is a actually a set, but it is a bit set. In future, when we add more options, once call pass multiple choices by simply ORing them. 
E.g. `capability = ALSurfaceCapability.TracInteraction | ALSurfaceCapability.TrackMutation`. That is super efficient and fast to handle. 
Is that what you were asking about? 
I could also choose an array of enums, but then it will be more verbose to pass, and slower to check.

packages/hyperion-autologging/src/ALFlowletManager.ts
reshadi: hmm, I didn't change the name because we will not have one of these per each capability. Only these two for now. In future, the other ones that I had in mind will be applied to both. 
What if I rename `region` to `nonInteractiveSurface` so we know this is only about interactivity, otherwise for both types we will track mutation.

reshadi:
@vsylee updated the PR with all the changes you requested.

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -206,7 +206,7 @@ function setupDomElementSurfaceAttribute(options: InitOptions): void {
 
 
 export function init(options: InitOptions): ALSurfaceHOC {
-  const { flowletManager, domSurfaceAttributeName = AUTO_LOGGING_SURFACE, domRegionAttributeName = AUTO_LOGGING_REGION } = options;
+  const { flowletManager, domSurfaceAttributeName = AUTO_LOGGING_SURFACE, domNonInteractiveSurfaceAttributeName: domRegionAttributeName = AUTO_LOGGING_NON_INTERACTIVE_SURFACe } = options;
```
vsylee, line 209:
```suggestion
  const { flowletManager, domSurfaceAttributeName = AUTO_LOGGING_SURFACE, domNonInteractiveSurfaceAttributeName = AUTO_LOGGING_NON_INTERACTIVE_SURFACE } = options;
```

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -231,7 +231,7 @@ export function init(options: InitOptions): ALSurfaceHOC {
     let domAttributeValue: string;
 
     const surfaceCtx = ALSurfaceContext.useALSurfaceContext();
-    const { surface: parentSurface, region: parentRegion } = surfaceCtx;
+    const { surface: parentSurface, nonInteractiveSurface: parentRegion } = surfaceCtx;
```
vsylee, line 234:
```suggestion
    const { surface: parentSurface, nonInteractiveSurface: parentNonInteractiveSurface } = surfaceCtx;
```

packages/hyperion-autologging/src/ALSurfaceConsts.ts
```diff
@@ -5,4 +5,4 @@
 'use strict';
 
 export const AUTO_LOGGING_SURFACE = "data-surface";
-export const AUTO_LOGGING_REGION = "data-region";
\ No newline at end of file
+export const AUTO_LOGGING_NON_INTERACTIVE_SURFACe = "data-non-int-surface";
```
vsylee, line 8:
```suggestion
export const AUTO_LOGGING_NON_INTERACTIVE_SURFACE = "data-non-int-surface";
```

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -248,14 +248,14 @@ export function init(options: InitOptions): ALSurfaceHOC {
...
-      fullRegionString = proxiedContext.region;
+      fullRegionString = proxiedContext.nonInteractiveSurface;
       domAttributeName = domSurfaceAttributeName
       domAttributeValue = fullSurfaceString;
     }
 
     const surfaceData: SurfaceData = {
       surface: fullSurfaceString,
-      region: fullRegionString,
+      nonInteractiveSurface: fullRegionString,
```
vsylee, line 258: If renaming, can we fully replace "region" with "nonInteractiveSurface"?

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -180,46 +206,87 @@ function setupDomElementSurfaceAttribute(options: InitOptions): void {
...
       __ext.flowlet = flowlet;
     }
 
-    const incomingSurfaceString = props.fullSurfaceString ?? '';
-    const isPassedSurface = incomingSurfaceString !== '';
-    let fullSurfaceString = incomingSurfaceString;
-    const { surface: parentSurface } = ALSurfaceContext.useALSurfaceContext();
-    if (!isPassedSurface) {
+    let fullSurfaceString: string;
+    let fullRegionString: string;
```
vsylee, line 229: At this point, wondering if also worth renaming this to "surface path" 

``` suggestion
    let surfacePath: string;
    let nonInteractiveSurfacePath: string;
```

packages/hyperion-autologging/src/ALSurface.ts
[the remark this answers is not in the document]
vsylee: Ahh yeah. Got it, that's pretty cool!

vsylee (approved):

packages/hyperion-autologging/src/ALSurface.ts
reshadi: Yes, I'll do that. Thanks for catching it. I relied on IDE doing all the renames.

PR #125, opened 2023-09-14
Trigger ui event even when propagation is stopped.

reshadi:
Ensuring that `al_ui_event` if fired in time so we can handle it before page unloads.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -4,6 +4,8 @@
 
 'use strict';
 import { Channel } from "@hyperion/hook/src/Channel";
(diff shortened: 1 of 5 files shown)


vsylee:
Overall looks good, but wondering if there is a way to test that emitting `al_ui_event` on `event.stopPropagation` will actually catch the event before navigating away (at least as a local repro)

reshadi:
> Overall looks good, but wondering if there is a way to test that emitting `al_ui_event` on `event.stopPropagation` will actually catch the event before navigating away (at least as a local repro)

I have tested it on the WWW side, generally we lack a good testing mechanism for the autologging package. I was thinking that at some point we should add jest tests to our test-react-app. 
I didn't want to import react anywhere in the hyperion-autologging to ensure we never accidentally bundle the whole react with it.

schwartzmx (approved):

PR #128, opened 2023-09-29
Trigger Flowlet

reshadi:
This serries of commits adds support for `triggerFlowlet` to accurately track cause / effect in the application.


packages/hyperion-autologging/src/ALNetworkPublisher.ts
```diff
@@ -226,7 +244,50 @@ function captureXHR(options: InitOptions): void {
     intercept.setVirtualPropertyValue<RequestInfo>(
       this,
       REQUEST_INFO_PROP_NAME,
-      { method, url: typeof url === 'string' ? url : url.href });
+      { method, url: typeof url === 'string' ? url : url.href }
+    );
+
+    const parentTriggerFlowlet = flowletManager.top()?.data.triggerFlowlet;
+    const triggerFlowlet = new flowletManager.flowletCtor(`xhr(method:${method},url:${url})`, parentTriggerFlowlet);
```
vsylee, line 251: Same issue with `url` here

vsylee (changes requested):
I was testing for the TCR case, and observed (at least for the limited local repros I did) that where ui_event_flowlet wasn't working, trigger_flowlet works, but where ui_event_flowlet *was* working, trigger_flowlet doesn't work :o Will work with you offline to walk through the specific examples

packages/hyperion-autologging/src/ALNetworkPublisher.ts
```diff
@@ -159,14 +166,20 @@ function captureFetch(options: InitOptions): void {
...
+        triggerFlowlet: flowlet?.data.triggerFlowlet,
         ...request,
         metadata: {},
       });
     } else {
       ephemeralRequestEvent = null;
     }
 
+    const parentTriggerFlowlet = flowletManager.top()?.data.triggerFlowlet;
+    const triggerFlowlet = new flowletManager.flowletCtor(`fetch(method:${request.method},url:${request.url})`, parentTriggerFlowlet);
```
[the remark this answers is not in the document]
reshadi: Good point! I actually left this here hesitantly. I wanted to also test how it will work with graphapi marking. 
If we decided to keep it, certainly only pathname makes more sense.

packages/hyperion-autologging/src/ALFlowletManager.ts
```diff
@@ -12,17 +12,37 @@ import { Flowlet } from "@hyperion/hyperion-flowlet/src/Flowlet";
...
+  implements IALFlowlet<DataType>
+{
+
+  toJSON() {
+    return {
+      ...flowletToJSON(this),
+      data: {
+        surface: this.data.surface,
+        uiEventFlowlet: flowletToJSON(this.data.uiEventFlowlet),
+        triggerFlowletN: flowletToJSON(this.data.triggerFlowlet),
```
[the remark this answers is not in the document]
reshadi: Oops! I think since I dropped the type of this, didn't get this error.

packages/hyperion-flowlet/src/TriggerFlowlet.ts
```diff
@@ -0,0 +1,30 @@
...
+//     super(name, parent);
+//     this.data.triggerFlowlet = this; // Ensure this flowlet report itself as the trigger.
+//   }
+// }
+
+export interface TriggerFlowlet extends Flowlet<FlowletDataType> { }
+
+export function setTriggerFlowlet(obj: TriggerFlowletPayload, triggerFlowlet: TriggerFlowlet): void {
+  triggerFlowlet.data.triggerFlowlet = triggerFlowlet;
+  obj.__ext_triggerFlowlet = triggerFlowlet;
```
vsylee, line 25: Elsewhere in Hyperion, I've seen you use `setVirtualPropertyValue`. Why not use that here as well?

reshadi: I started with that, but then I realized we want to use this same function to also assign triggerFlowlet to normal objects, such as action payload, and graphapi params. 
`setVirtualPropertyValue` requires the object to be interceptable. 
So, for now I'm hiding the actual implementation behind these two function (set/getTriggerFlowlet). If we decided to allow normal objects be intercepted (e.g. add IObject to core), then we can switch the implementation.

vsylee:
Could you update this PR by resolving the merge conflicts with https://github.com/facebook/hyperion/pull/130?

reshadi:
> Could you update this PR by resolving the merge conflicts with #130?

Done

vsylee (approved):

PR #130, opened 2023-10-10
Add ref props to surface

danikawidjaja:
Adding an optional nodeRef prop to hyperion surface so we can track mutations with that ref instead of wrapping it with an additional span. This is because wrapping it with an additional span can obstruct/interfere with browser visibility logic. 

## Tests
Screenshot 2023-10-10 at 8 46 13 AM

Build and run all tests pass: https://www.internalfb.com/phabricator/paste/view/P850540839


vsylee (approved):
LGTM, thanks for fixing this and testing!

PR #132, opened 2023-10-17
Add a heartbeat for end of page.

reshadi:
We already fire a heartbeat when page starts, this commit also adds a heartbeat for when page is about to unload.

This way applications can ensure all events are logged by ensuring each session has heartbeat of type START and STOP


vsylee (commented):
Is it also possible to add tests for the page unload case?

reshadi:
> Is it also possible to add tests for the page unload case?

I tried, but couldn't think of something. I added some concole.log and debugger lines to make sure it works. Our SPA react app does not quite allow that. Any suggestions?

vsylee:
> > Is it also possible to add tests for the page unload case?
> 
> I tried, but couldn't think of something. I added some concole.log and debugger lines to make sure it works. Our SPA react app does not quite allow that. Any suggestions?

Got it. I also saw you were able to test locally in WWW, so will accept for now.

vsylee (approved):

PR #134, opened 2023-10-20
[donotmerge] Allow disabling of surface proxy

schwartzmx:
As a last resort if needed.


packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -351,7 +353,9 @@ export function init(options: InitOptions): ALSurfaceHOC {
     return result;
   }
 
-  SurfaceProxy.init({ ...options, surfaceComponent: Surface });
+  if (!options.disableSurfaceProxy === true) {
```
reshadi, line 356: Even if we wanted to do this, the code should sit inside of the SurfaceProxy.init itself to ensure every path is handled.

PR #137, opened 2023-10-31
Add flowlet group for mouse events

vsylee:
Introducing a concept of grouping UI events that can happen in a sequence, e.g. mouseover -> click -> mouseout, to enable correlating events triggered by the same set of mouse events. Starting with coverage for the mouseover, click, and mouseout events, based on the mouse events enabled for tracking from WWW.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -111,6 +111,12 @@ const shouldPushPopFlowlet = (event: Event) => event.bubbles && event.isTrusted;
 
 const activeUIEventFlowlets = new Map<UIEventConfig['eventName'], IALFlowlet>();
 
+enum UIEventGroup {
+  MOUSE = 'mouse',
+  // KEY = 'key',
+}
+const groupFlowlets = new Map<EventTarget, {groupFlowlet: IALFlowlet, type: UIEventGroup}>();
+
 const uiEventFlowletManager = new ALFlowletManager();
 
 function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEventConfig<DocumentEventMap>, eventName: T, event: DocumentEventMap[T]): CommonEventData | null {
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -196,6 +200,21 @@ export function publish(options: InitOptions): void {
       const topFlowlet = flowletManager.top();
       let flowlet = topFlowlet ?? defaultTopFlowlet; // We want to ensure flowlet is always assigned
       if (shouldPushPopFlowlet(event)) {
+        let groupFlowlet;
+        if (event.target != null) {
+          const targetElement = event.target;
```
reshadi, line 205: this is also available in the `uiEventData` above.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -196,6 +200,21 @@ export function publish(options: InitOptions): void {
...
+            const type = UIEventGroup.MOUSE;
+            groupFlowlet = new flowletManager.flowletCtor(`${type}(element:${autoLoggingID})`);
+            groupFlowlets.set(targetElement, {groupFlowlet, type});
+          } else if (eventName === 'mouseout' || eventName === 'click') {
+            const groupFlowletInfo = groupFlowlets.get(targetElement);
+            if (groupFlowletInfo != null && groupFlowletInfo.type === UIEventGroup.MOUSE) {
+              groupFlowlet = groupFlowletInfo.groupFlowlet;
+            }
+          }
+        }
```
reshadi, line 216: This logic is ok, but I think a better approach is to have a static map from `eventName` to `(autoLoggingID: string)=>IALFlowlet | undefined` outside of the event handler, and then just read the `groupFlowletGetter` function outside and then run it here. 
This way, as we add more and more events to the list, it is easy to extend it and cost remains constant.

reshadi: Based on our offline conversation, I gave this a try as part of this [PR](https://github.com/facebook/hyperion/pull/138/files#diff-062df1927fb75ce28648c7b3680f678c0be0efa6d2e4c7c406a710b054550438), since I needed something to control general trigger flowlet assignment.

schwartzmx:
Think we agreed on closing this one in favor of #138 including this functionality

PR #138, opened 2023-11-02
Triggerflowlet update

reshadi:
Fixed trigger flowlet for network_response as well as event handlers.


packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -2,20 +2,27 @@
...
 import * as IReact from "@hyperion/hyperion-react/src/IReact";
 import * as IReactComponent from "@hyperion/hyperion-react/src/IReactComponent";
-import { IALFlowlet, ALFlowletManager } from "./ALFlowletManager";
+import TestAndSet from "@hyperion/hyperion-util/src/TestAndSet";
+import performanceAbsoluteNow from "@hyperion/hyperion-util/src/performanceAbsoluteNow";
+import { ALFlowletManager, IALFlowlet } from "./ALFlowletManager";
+import { isTrackedEvent } from "./ALInteractableDOMElement";
 import { ALChannelSurfaceEvent } from "./ALSurface";
 import { ALSurfaceContext, ALSurfaceContextFilledValue, useALSurfaceContext } from "./ALSurfaceContext";
+import * as ALUIEVentGroupPublishers from "./ALUIEventGroupPublisher";
```
schwartzmx, line 23:
```suggestion
import * as ALUIEventGroupPublishers from "./ALUIEventGroupPublisher";
```

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -2,20 +2,27 @@
...
+'use strict';
 import type * as Types from "@hyperion/hyperion-util/src/Types";
 import type * as React from 'react';
 
 import { Channel } from "@hyperion/hook/src/Channel";
 import { getFunctionInterceptor } from "@hyperion/hyperion-core/src/FunctionInterceptor";
-import { TriggerFlowlet } from "@hyperion/hyperion-flowlet/src/TriggerFlowlet";
+import { getVirtualPropertyValue, setVirtualPropertyValue } from "@hyperion/hyperion-core/src/intercept";
+import * as IEventTarget from "@hyperion/hyperion-dom/src/IEventTarget";
+import * as FLowlet from "@hyperion/hyperion-flowlet/src/Flowlet";
```
schwartzmx, line 13:
```suggestion
import * as Flowlet from "@hyperion/hyperion-flowlet/src/Flowlet";
```

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -86,6 +101,38 @@ export function init(options: InitOptions) {
...
+    }
+
+    if (!isTrackedEvent(eventType) && this instanceof Element && !getVirtualPropertyValue(this, IS_TRIGGER_FLOWLET_SETUP_PROP)) {
+      setVirtualPropertyValue(this, IS_TRIGGER_FLOWLET_SETUP_PROP, true);
+      /**
+       * We could potentially intercept the callback itself and add the following
+       * but that could add too much overhead to every callback.
+       */
+      // const topTriggerFlowlet = flowletManager.top()?.data.triggerFlowlet;
+      IEventTarget.addEventListener.getOriginal().call(
```
schwartzmx, line 118: For my understanding what is this doing exactly?  We are inside of an `onArgsObserverAdd` for the same `addEventListener` that we are intercepting but then doing a call on the original function that was intercepted?

packages/hyperion-flowlet/test/FlowletManager.test.ts
```diff
@@ -190,4 +190,21 @@ describe("test FlowletManager", () => {
...
+      expect(manager.top()?.getFullName()).toBe('/main/foo');
+    };
+
+    const foo = () => {
+      bar();
+    }
+
+    const markedFoo = manager.mark(foo, 'foo');
+
+    markedFoo();
```
schwartzmx, line 208: So here `markedFoo` would push a new `foo` flowlet when called?

schwartzmx (commented):
Overall I think mostly everything made sense,  just a couple questions for my understanding

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
reshadi: I'm glad you noticed this! :).
So, basically we catch application code calling `addEventListener` here to add some event. We then want to do one extra event registration of our own, the very first time such event handler is installed. Here if we call the normal `addEventListener`, it will again bring us back to this very interception and we go into an infinite loop! 
We do know that the goal of this call back is to update the triggerFlowlet and nothing else matters. So, we can bypass all of the other logic by calling the original function. 

I'll add this comment to the code itself in an update.

schwartzmx (approved):

PR #140, opened 2023-11-09
Fix arrow function component visitors during interception

reshadi:
Function components that are created via arraow (=>) functions won't have `.prototype` value. The code already had a bug that was checking this value before trying to get its prototype! This would throw at runtime but someone react was always hiding this error!

### DIFF
--- a/packages/hyperion-react/src/IReactComponent.ts
+++ b/packages/hyperion-react/src/IReactComponent.ts
@@ -226,30 +226,28 @@ export function init(options: InitOptions): void {
              *  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/prototype#:~:text=prototype%20property%2C%20by%20default
            */
 
-          // $FlowIgnore[prop-missing]
           const classComponentParentClass = component.prototype;
+          let classComponentParentClassParent;
 
           if (
-            classComponentParentClass instanceof ReactModule.Component ||
-            typeof classComponentParentClass?.render === 'function' || // possibly created via React.createClass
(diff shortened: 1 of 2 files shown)


packages/hyperion-react/src/IReactComponent.ts
```diff
@@ -226,30 +226,28 @@ export function init(options: InitOptions): void {
...
-            classComponentParentClass instanceof ReactModule.Component ||
-            typeof classComponentParentClass?.render === 'function' || // possibly created via React.createClass
-            Object.getPrototypeOf(Object.getPrototypeOf(classComponentParentClass)) || // not a plain function, may be with some other lifecycle methods. 
-            classComponentParentClass === ReactModule.Component.prototype // in case buggy code didn't properly inherit from ReactModule.Component
+            classComponentParentClass && (
+              classComponentParentClass instanceof ReactModule.Component ||
+              typeof classComponentParentClass.render === 'function' || // possibly created via React.createClass
+              (
+                (classComponentParentClassParent = Object.getPrototypeOf(classComponentParentClass)) &&
+                Object.getPrototypeOf(classComponentParentClassParent)
```
vsylee, line 238: Was the silent error throwing when calling `Object.getPrototypeOf` on null `Object.getPrototypeOf(classComponentParentClass)`?

vsylee (approved):

packages/hyperion-react/src/IReactComponent.ts
reshadi: Yes, exactly. There was already check for the null for other lines, except that one case.

PR #143, opened 2023-11-11
Handle long chaning of flowlets

reshadi:
In some cases, we might have a very long chain of flowlets (e.g. recursive setTimeout calls).

When we calculate the fullName of such cases, we might run into stack overflow.

This commit changes the algorithm to limit the length of the chain that we walk.

### DIFF
--- a/packages/hyperion-flowlet/test/Flowlet.test.ts
+++ b/packages/hyperion-flowlet/test/Flowlet.test.ts
@@ -8,6 +8,7 @@ import { Flowlet } from "../src/Flowlet";
 describe("test Flowlet", () => {
   test("test Flowlet methods", () => {
     const f1 = new Flowlet<{
+      triggerFlowlet: any,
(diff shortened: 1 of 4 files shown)


packages/hyperion-flowlet/test/Flowlet.test.ts
```diff
@@ -23,4 +24,15 @@ describe("test Flowlet", () => {
...
   });
 
+  test("long flowlet chain", () => {
+    let flowlet = new Flowlet<{}>('f1');
+    for (let i = 0; i < 100000; ++i) {
+      flowlet = flowlet.fork('f');
+    }
+
+    const name = flowlet.getFullName();
+    expect(name.length).toBeGreaterThan(1000);
```
schwartzmx, line 34: Should this check that it is less than the max?

schwartzmx (approved):

PR #145, opened 2023-11-21
When wrapping a Portal'd SVG element, wrap in `<g>` not `<span>`

spencer516:
When using autologger and a React portal, the element ends up getting wrapped with a `<span>`. If this is an SVG element, however, this will break the SVG as a `<span>` is not a valid element in the SVG namespace. 

This PR detects if the element that is being wrapped is an SVG element and changes the wrapping element to a `<g>`

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -327,6 +327,9 @@ export function init(options: InitOptions): ALSurfaceHOC {
 
     // flowlet.data.surface = surfacePath;
     let children = props.children;
+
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging-visualizer/src/component/ElementTextTooltip.react.js
reshadi, line 1: Make sure you exclude this file from the final commit. There should a .gitignore already in the project that prevents .js files being part of commit. I wonder why that is not working.

reshadi (approved):
Thanks for fixing this issue.

reshadi (approved):

PR #146, opened 2023-11-29
Add support for dynamic flowlet naming in `FlowletManager.mark`

akunieda:
Sometimes it's useful to name the flowlets created using `FlowletManager.mark` dynamically.


reshadi (approved):
This is great! Thanks for fixing and adding test.

PR #147, opened 2023-11-30
Promisemergetrigger

reshadi:
Handle merging of trigger flowlets for promise.


packages/hyperion-flowlet/src/FlowletWrappers.ts
```diff
@@ -181,14 +181,56 @@ export function initFlowletTrackers(flowletManager: FlowletManager) {
...
+          if (arg instanceof Promise) {
+            const triggerFlowlet = getTriggerFlowlet(arg);
+            if (triggerFlowlet) {
+              triggerFlowlets.push(triggerFlowlet);
+            }
+          }
+        }
+
+        if (triggerFlowlets.length > 0) {
+          const flowletName = `${fi.name}{${triggerFlowlets.map(f => f.id).join(",")}}`;
```
vsylee, line 221: Maybe also append `Promise.` at the beginning- I think people will not know that it's related to Promise when they just see `all` for instance

Also wondering if we should use `()` instead of `{}`, since we've typically been using `()` to add extra info for other flowlets

vsylee (approved):

packages/hyperion-flowlet/src/FlowletWrappers.ts
reshadi: Good suggestion, updated.

PR #148, opened 2023-12-05
Rename flowlets to always include ID in the name.

reshadi:
In many cases, we do need to distinguish between instances of the flowlets. So, now the inclusion of id in the name is by default enabled.

We can now remove timestamp from the flowlet names and end up with shorter names.

Also, replaced use of `,` in the names with `&`, and explicitly renamed the root flowlet to `/`. This way, the flowlet will look like a URI and is easier to copy/paste in other tools.

Consumers of the flowlet can also add a `schema` before the name before logging to embed the meaning the name, e.g.
`trigger://:1/mounse:2/click:3/`.

### DIFF
--- a/packages/hyperion-flowlet/src/Flowlet.ts
+++ b/packages/hyperion-flowlet/src/Flowlet.ts
@@ -28,7 +28,7 @@ export class Flowlet<T extends FlowletDataType = FlowletDataType> {
     onFlowletInit.call<T>(this);
   }
 
-  getFullName(includeId:boolean = false): string {
+  getFullName(): string {
     if (!this._fullName) {
       // The following is easiest approach, but does not work for very long chains!
       // this._fullName = `${this.parent?.getFullName() ?? ""}/${this.name}`;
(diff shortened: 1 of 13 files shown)


packages/hyperion-flowlet/src/Flowlet.ts
```diff
@@ -46,10 +46,7 @@ export class Flowlet<T extends FlowletDataType = FlowletDataType> {
       let prefix: string = f ? (f._fullName ?? '...') : '';
       for (let j = rawFlowlets.length-1; j>=0; --j){
         let flowlet = rawFlowlets[j];
-        flowlet._fullName = prefix = prefix + "/" + flowlet.name;
-        if (includeId){
-          flowlet._fullName += ":" + flowlet.id;
-        }
+        flowlet._fullName = prefix = prefix + "/" + flowlet.name + ":" + flowlet.id;
```
vsylee, line 49: With these changes, we are now using `:` to both append the flowlet id and add flowlet data details, e.g. `xhr(method:${method}&url:${request.uri.pathname})` for network responses. This flowlet's full name would then look something like `/xhr(method:POST&url:/api/graphql/):170287`. Wondering if we should use a different delimiter for the flowlet id (maybe `_`?) to make it easier to parse, e.g. in TA

vsylee (approved):

packages/hyperion-flowlet/src/Flowlet.ts
reshadi: Great observation, since I replaced the `,` with `&` to be more URI like format, one option is to use `=` instead of `:` for flowlet data details. Then your example will look like `/xhr(method=POST&url=/api/graphql/):170287`, what do you think?

PR #150, opened 2023-12-08
Split trigger flowlet logic

reshadi:
To measure the efficacy and perf impact of different parts of the trigger flowlet, split the logic to smaller pieces controlled by different flags.

### DIFF
--- a/packages/hyperion-autologging/src/ALTriggerFlowlet.ts
+++ b/packages/hyperion-autologging/src/ALTriggerFlowlet.ts
@@ -145,7 +145,7 @@ export function init(options: InitOptions) {
     }
   });
 
-  // Temporarily disable this until we can re-enable react-flowlet logic bellow
+  // Temporarily disable this until we can re-enable react-flowlet logic below
   if (options.enablePerSurfaceTracking) {
     const ALSurfaceContextDataMap = new Map<ALSurfaceContextFilledValue['surface'], ALSurfaceContextFilledValue['flowlet']>();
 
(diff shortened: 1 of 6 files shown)


packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -172,6 +145,55 @@ export function init(options: InitOptions) {
     }
   });
 
+  // Temporarily disable this until we can re-enable react-flowlet logic bellow
```
vsylee, line 148: nit

```suggestion
  // Temporarily disable this until we can re-enable react-flowlet logic below
```

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -172,6 +145,55 @@ export function init(options: InitOptions) {
...
+    });
+    channel.addListener('al_surface_unmount', event => {
+      ALSurfaceContextDataMap.delete(event.surface);
+      activeRootFlowlets.delete(event.flowlet);
+    });
+  }
+
+  if (options.enableFlowletConstructorTracking) {
+    //@ts-expect-error
+    flowletManager.root.data.isRootet = true;
```
vsylee, line 182: Is this field supposed to be `isRooted`? Could you help explain a bit about what this is tracking?

vsylee (approved):

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
reshadi: Yes, thanks for catching the typo. I added the full description as comment in the code. Adding it here for reference:
```
    /**
     * To know if root of a flowlet is added to list, we would need
     * to walk up the .parent chain to find the root.
     * Since we know most of the flowlets will be created from the
     * FlowletManager.root, we use a the following trick:
     * We add a known field the root's data and check chat in children's
     * data (data fields inherit from each other).
     * If we don't find the field, we have two options:
     * 1- Now walk up the .parent chain (slow path)
     * 2- Consider this case error and fix it in the code.
     * For now, we choose 2 since there is no real legit case for a
     * 'rootless' flowlet!
     */
```

PR #151, opened 2023-12-11
Use window.crypto.getRandomValues to generate guids

williewillus:
Internal: See T164417895

### DIFF
--- a/packages/hyperion-util/src/guid.ts
+++ b/packages/hyperion-util/src/guid.ts
@@ -8,5 +8,5 @@
 export type GUID = string;
 
 export function guid(): GUID {
-  return 'f' + (Math.random() * (1 << 30)).toString(16).replace('.', '');
+  return 'f' + (window.crypto.getRandomValues(new Uint32Array(1))[0]).toString(16).replace('.', '');
 }


packages/hyperion-util/src/guid.ts
```diff
@@ -8,5 +8,5 @@
 export type GUID = string;
 
 export function guid(): GUID {
-  return 'f' + (Math.random() * (1 << 30)).toString(16).replace('.', '');
+  return 'f' + (window.crypto.getRandomValues(new Uint32Array(1))[0]).toString(16).replace('.', '');
```
reshadi, line 11: Thanks for updating this. This code can run outside of main thread (e.g. in WebWorker). To be safe you should remove the `window` and instead use `globalThis`. It is also safer to use feature detection and first check to make sure `globalThis.crypto` is available.

e.g.
```
const getRandom : ()=>number = globalThis.cryto 
  ? ()=> globalThis.crypto.getRandomValues(new Uint32Array(1))[0]
  : ()=> Math.random()
;
export function guid(): GUID {
  return 'f' + (getRandom()).toString(16).replace('.', '');
}
```

reshadi (changes requested):

reshadi:
This change is more accurately incorporated into: https://github.com/facebook/hyperion/pull/211.

PR #152, opened 2024-01-09
Make sure pageload flowlet has a root

reshadi:
An assersion was firing a lot because the `pageload` trigger flowlet did not have root. Others do.

This fixes the issue.

### DIFF
--- a/packages/hyperion-autologging/src/ALTriggerFlowlet.ts
+++ b/packages/hyperion-autologging/src/ALTriggerFlowlet.ts
@@ -52,7 +52,7 @@ export function init(options: InitOptions) {
 
   const { channel, flowletManager } = options;
 
-  let currTriggerFlowlet = new flowletManager.flowletCtor('pageload');
+  let currTriggerFlowlet = new flowletManager.flowletCtor('pageload', flowletManager.root);
   currTriggerFlowlet.data.triggerFlowlet = currTriggerFlowlet;
 
   const activeRootFlowlets = new class {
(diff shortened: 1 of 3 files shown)


schwartzmx (approved):

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
```diff
@@ -52,7 +52,7 @@ export function init(options: InitOptions) {
 
   const { channel, flowletManager } = options;
 
-  let currTriggerFlowlet = new flowletManager.flowletCtor('pageload');
+  let currTriggerFlowlet = new flowletManager.flowletCtor('pageload', flowletManager.root);
```
vsylee, line 55: `currTriggerFlowlet` did have a root flowlet, just not the "right" root flowlet that the "unexpected root flowlet" check was looking for, right?

vsylee (approved):

packages/hyperion-autologging/src/ALTriggerFlowlet.ts
reshadi: `currTriggerFlowlet` was the root of many other triggers (e.g. actions). But it itself was not forking the root flowlet. 
The assumption in the code is that for known cases, we fork the root so that we can assign values just to the root and everyone automatically inherits it. 
We already do this for UI events, but had missed doing it here as well.

vsylee: Then should we also update the `FlowletManager` constructor to default to `flowletManager.root` if `parent` is not defined in the constructor arguments? I'm concerned that we may have other gaps or will miss this again when creating new flowlets

PR #153, opened 2024-01-29
Simplify DOM surface attribute name options

vsylee:
Proposing to simplify setting the DOM surface attribute name by removing the ability to pass these fields through options, i.e. to use the `ALSurfaceConsts` values from Hyperion and not allow overwriting from WWW. One reason for this is to simplify initialization of AL code and make it easier to test specific pieces in WWW environments (e.g. D53191534).

### DIFF
--- a/packages/hyperion-autologging-visualizer/src/index.ts
+++ b/packages/hyperion-autologging-visualizer/src/index.ts
@@ -9,7 +9,7 @@ import { Channel } from "@hyperion/hook/src/Channel";
 import * as ElementTextTooltip from "./component/ElementTextTooltip.react";
 
 export type InitOptions = Types.Options<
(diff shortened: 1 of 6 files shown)


reshadi:
I'm ok with this change, but we have to ensure that whatever depends on the old name is updated. 
Also, for all of these cases, we had made sure that these attributes be optional, meaning, if you don't pass them, the effect will be the same as the change you applied. So, I'm not sure how it helps?

vsylee:
> I'm ok with this change, but we have to ensure that whatever depends on the old name is updated.

Yup, check out D53194580

> we had made sure that these attributes be optional

`domSurfaceAttributeName` was actually a required option. `domNonInteractiveSurfaceAttributeName` wasn't exposed, but I modified it for consistency. There's still also `domFlowletAttributeName`, but that one is currently set to depend on WWW environment variables.

reshadi (approved):

PR #156, opened 2024-02-08
Expose ALElementInfo static method for resolving react component on d…

schwartzmx:
…emand, including optional ability to stop fiber walking

This is a slightly different route than https://github.com/facebook/hyperion/pull/154
But a quicker alternative than trying to deal with the ElementInfo cache problem described in our meeting and witnessed in https://github.com/facebook/hyperion/compare/main...schwartzmx:hyperion:bail-traversal with trying to make this configurable per event.

### DIFF
--- a/packages/hyperion-autologging/src/ALElementInfo.ts
+++ b/packages/hyperion-autologging/src/ALElementInfo.ts
@@ -11,6 +11,7 @@ import * as IElement from "@hyperion/hyperion-dom/src/IElement";
 
 const AL_ELEMENT_INFO_PROPNAME = '__alInfo';
 const AUTO_LOGGING_COMPONENT_TYPE = 'data-auto-logging-component-type';
+export type BailTraversalFunc = (foundReactComponent: boolean, depth: number) => boolean;
 
 export default class ALElementInfo {
   element: Element;
@@ -44,6 +45,10 @@ export default class ALElementInfo {
 
   static getOrCreate(element: Element): ALElementInfo {
     return ALElementInfo.get(element) ?? new ALElementInfo(element);
+  }
+
+  static getReactComponentData(element: Element, bailTraversalFunc?: BailTraversalFunc): ReactComponentData | null {
+    return getReactComponentData_THIS_CAN_BREAK(element, bailTraversalFunc);
   }
 
   getReactComponentData(): ReactComponentData | null {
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -46,6 +47,10 @@ export default class ALElementInfo {
     return ALElementInfo.get(element) ?? new ALElementInfo(element);
   }
 
+  static getReactComponentData(element: Element, bailTraversalFunc?: BailTraversalFunc): ReactComponentData | null {
+    return getReactComponentData_THIS_CAN_BREAK(element, bailTraversalFunc);
```
reshadi, line 51: to make sure this function never gets too expensive, I suggest, here first try to use the cache (`.get()`) and then call the rest. 

Also, the `getReactComponentData_THIS_CAN_BREAK` can return `{name, stack, isTruncated}` to keep track of weather the bail function was called. If not, then it can just update the cache. 

You can add this logic later though.

schwartzmx: The caching occurs in the constructor so we'd maybe need to allow the bailing to be done up front?  If we go the route of trying to conditionally set it or not set it if bailed

packages/hyperion-autologging/src/ALElementInfo.ts
```diff
@@ -46,6 +47,10 @@ export default class ALElementInfo {
     return ALElementInfo.get(element) ?? new ALElementInfo(element);
   }
 
+  static getReactComponentData(element: Element, bailTraversalFunc?: BailTraversalFunc): ReactComponentData | null {
+    return ALElementInfo.get(element)?.getReactComponentData() ?? getReactComponentData_THIS_CAN_BREAK(element, bailTraversalFunc);
+  }
```
reshadi, line 52: great, now all you need to know if first part didn't work, and second part didn't return `isTruncated`, then you can call `.getOrCreate()` function and update the `this.reactComponentData` field for it. (see bellow).

packages/hyperion-autologging/src/ALElementInfo.ts
reshadi: see my inline response. I think you can now easily update the cached data. But that's optional for next level of efficiency.

PR #157, opened 2024-02-26
Update list of React symbols based on recent changes.

reshadi:
Recently React added one more symbol for handling Context.Consumer. That started to trigger one of the asserts in the code in the dev mode. This diff integrates that new changes.


vsylee (approved):
Thanks for the fix!

PR #158, opened 2024-03-13
Handle react inserted interactable element with no fiber

schwartzmx:
Handle cases where react inserts a node with listeners attached, but lacking a _reactFiber.  In these cases we can't traverse the fiber, since it's not available on the node that is marked interactable.

This change checks for this case and continues fiber walking from the parent element.

### DIFF
--- a/packages/hyperion-autologging/src/ALReactUtils.ts
+++ b/packages/hyperion-autologging/src/ALReactUtils.ts
@@ -46,9 +46,24 @@ type ReactInternalFiberType = Readonly<{
 }>;
 
 const getReactInternalFiber = (element: Element): ReactInternalFiber | null => {
-  const key = Object.keys(element).find(key => key.startsWith('__reactFiber$'));
+  let fiberKey: string | null = null;
+  let listeningKeyFound = false;
+  // Iterate the keys of the element, looking for react specific props
+  // If this is a reactListening node, then fiber will not be available, attempt to grab fiber from the parent element.
+  for (const key of Object.keys(element)) {
+    if (key.startsWith('__reactFiber$')) {
+      fiberKey = key;
+      break;
+      // Note this does not have the $ suffix, or begin with two _
+    } else if (key.startsWith('_reactListening')) {
+      listeningKeyFound = true;
+    }
+  }
+  if (fiberKey == null && listeningKeyFound) {
+    return getReactInternalFiber(element.parentElement as Element);
+  }
   const el = element as { [k: string]: any };
-  return key != null ? el[key] : null;
+  return fiberKey != null ? el[fiberKey] : null;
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALReactUtils.ts
```diff
@@ -46,9 +46,24 @@ type ReactInternalFiberType = Readonly<{
...
+    if (key.startsWith('__reactFiber$')) {
+      fiberKey = key;
+      break;
+      // Note this does not have the $ suffix, or begin with two _
+    } else if (key.startsWith('_reactListening')) {
+      listeningKeyFound = true;
+    }
+  }
+  if (fiberKey == null && listeningKeyFound) {
+    return getReactInternalFiber(element.parentElement as Element);
```
reshadi, line 63: Shouldn't you check if `parentElement` is not null? I think the `as` cast is possibly hiding the issue.

packages/hyperion-autologging/src/ALReactUtils.ts
```diff
@@ -46,9 +46,24 @@ type ReactInternalFiberType = Readonly<{
 }>;
 
 const getReactInternalFiber = (element: Element): ReactInternalFiber | null => {
-  const key = Object.keys(element).find(key => key.startsWith('__reactFiber$'));
+  let fiberKey: string | null = null;
+  let listeningKeyFound = false;
+  // Iterate the keys of the element, looking for react specific props
+  // If this is a reactListening node, then fiber will not be available, attempt to grab fiber from the parent element.
+  for (const key of Object.keys(element)) {
+    if (key.startsWith('__reactFiber$')) {
```
reshadi, line 54: Btw, I think we can make this much faster by finding the hash value that is used for the application. I think that value is fixed during lifetime of the application, e.g. `__reactFiber$sftyze5gh5`. So, may be if we don't know the full name, search for it once, but if we cannot find that, then do the search in the loop.

packages/hyperion-autologging/src/ALReactUtils.ts
schwartzmx: Yeah,  good call.  I don't think it was hiding anything but it should probably be checked,  this is new code.  Wouldn't it only be null for `window.document` or are there other scenarios?  Regardless I pushed a check for this

packages/hyperion-autologging/src/ALReactUtils.ts
schwartzmx: This is a good idea!  And yes that seems to be the case,  makes this function a little more involved but I think any perf wins will be worth it.   It seems to work as expected,  please see latest push

reshadi (approved):

PR #159, opened 2024-03-14
Add API for allowing disabling/enabling of UIPublisher handlers by event

schwartzmx:
Use a map to maintain handlers as they are added,  and for reference when re-enabling the handlers.

The two exported functions are very similar,  I'm not opposed to combining them into a single call if there's any suggestions.

Tested that the APIs work as expected in www
https://pxl.cl/4vGxg

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -84,30 +84,43 @@ function ignoreInteractiveElement(node: HTMLElement) {
 }
 
 // Keep track of event handlers
-export const UIEventHandlers = new Map<UIEventConfig['eventName'] | string, {
+export type TrackEventHandlerConfig = Readonly<{
   captureHandler: (event: Event) => void,
   bubbleHandler: (event: Event) => void,
   active: boolean,
 }>;
+export const UIEventHandlers = new Map<UIEventConfig['eventName'] | string, TrackEventHandlerConfig>;
 
 export function disableUIEventHandlers(eventName: UIEventConfig['eventName']): void {
   const handlerConfig = UIEventHandlers.get(eventName);
   if (handlerConfig?.active === true) {
     window.document.removeEventListener(eventName, handlerConfig.captureHandler, true);
     window.document.removeEventListener(eventName, handlerConfig.bubbleHandler, false);
-    handlerConfig.active = false;
+    UIEventHandlers.set(eventName, {
+      ...handlerConfig,
+      active: false,
+    });
   }
 }
 
-export function enableUIEventHandlers(eventName: UIEventConfig['eventName']): void {
-  const handlerConfig = UIEventHandlers.get(eventName);
+export function enableUIEventHandlers(eventName: UIEventConfig['eventName'], eventHandlerConfig?: TrackEventHandlerConfig | undefined): void {
+  let handlerConfig = UIEventHandlers.get(eventName) ?? eventHandlerConfig;
+  // Incoming config
+  if (eventHandlerConfig != null) {
+    // Disable the existing handlers if present, before installing the new ones
+    disableUIEventHandlers(eventName);
+    handlerConfig = eventHandlerConfig;
+  }
   if (handlerConfig?.active === false) {
     // Install interactable attribute handlers
     installHandlers();
     // Install event handlers
     window.document.addEventListener(eventName, handlerConfig.captureHandler, true);
     window.document.addEventListener(eventName, handlerConfig.bubbleHandler, false);
-    handlerConfig.active = true;
+    UIEventHandlers.set(eventName, {
+      ...handlerConfig,
+      active: true,
+    });
   }
 }
 
@@ -138,7 +151,7 @@ let installHandlers = () => {
 
   IReactComponent.onReactDOMElement.add((_component, props: ReactComponentObjectProps) => {
     if (props != null) {
-      Array.from(UIEventHandlers.keys()).forEach(event => {
+      UIEventHandlers.forEach((_, event) => {
         if (props[SYNTHETIC_EVENT_HANDLER_MAP[event]] != null) {
           props[eventHandlerTrackerAttribute(event)] = '1';
         }
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -7,6 +7,7 @@ import * as IEventTarget from "@hyperion/hyperion-dom/src/IEventTarget";
 import { ReactComponentObjectProps } from "@hyperion/hyperion-react/src/IReact";
 import * as IReactComponent from "@hyperion/hyperion-react/src/IReactComponent";
 import type * as Types from "@hyperion/hyperion-util/src/Types";
+import { UIEventConfig } from "./ALUIEventPublisher";
 
 'use strict';
 
@@ -49,7 +50,7 @@ function eventHandlerTrackerAttribute(eventName: string): string {
 
 export function getInteractable(
   node: EventTarget | null,
-  eventName: string,
+  eventName: UIEventConfig['eventName'],
   returnInteractableNode: boolean = false,
 ): HTMLElement | null {
   // https://www.w3.org/TR/2011/WD-html5-20110525/interactive-elements.html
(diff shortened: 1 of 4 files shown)


vsylee (approved):
Seems like a reasonable approach to get this working for now, cc @reshadi in case you had any specific implementation in mind

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -158,6 +158,38 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
...
+  bubbleHandler: (event: Event) => void,
+  active: boolean,
+}>;
+
+export function disableEventHandlers(eventName: UIEventConfig['eventName']): void {
+  const handlerConfig = uiEventHandlers.get(eventName);
+  if (handlerConfig?.active === true) {
+    window.document.removeEventListener(eventName, handlerConfig.captureHandler, true);
+    window.document.removeEventListener(eventName, handlerConfig.bubbleHandler, false);
+    uiEventHandlers.set(eventName, {
```
reshadi, line 174: although this is correct, but create extra objects, you could also just set active to false. Unless, you define the data in the map as readonly to be more safe.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -229,7 +261,9 @@ export function publish(options: InitOptions): void {
       intercept(event); // making sure we can track changes to the Event object
       setTriggerFlowlet(event, flowlet);
       channel.emit('al_ui_event_capture', eventData);
-    },
+    };
+    window.document.addEventListener(eventName,
```
reshadi, line 265: better to call the `enableEventHandlers` to make sure we always use exact same mechanism.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -245,43 +279,51 @@ export function publish(options: InitOptions): void {
...
+      }
+    };
+    window.document.addEventListener(
+      eventName,
+      bubbleHandler,
       false, // useCapture
     );
+
+    // Set our registered handlers
+    uiEventHandlers.set(eventName, {
```
reshadi, line 322: you could set this as `active: false`, and then call `enableEventHandlers`

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -245,43 +279,51 @@ export function publish(options: InitOptions): void {
...
-        }
-      },
+      let flowlet: IALFlowlet | undefined;
+      if (shouldPushPopFlowlet(event) && (flowlet = activeUIEventFlowlets.get(eventName)) != null) {
+        flowletManager.pop(flowlet);
+        activeUIEventFlowlets.delete(eventName);
+        uiEventFlowletManager.pop(flowlet);
+      }
+    };
+    window.document.addEventListener(
```
reshadi, line 315: same as above

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -158,6 +158,38 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
...
 
+
+// Keep track of event handlers
+const uiEventHandlers = new Map<UIEventConfig['eventName'], {
+  captureHandler: (event: Event) => void,
+  bubbleHandler: (event: Event) => void,
+  active: boolean,
+}>;
+
+export function disableEventHandlers(eventName: UIEventConfig['eventName']): void {
```
reshadi, line 169: since you're exporting these methods, call them `{enable|disable}UIEventHandlers`

src/index.ts
```diff
@@ -62,3 +62,4 @@ export * as AutoLogging from "@hyperion/hyperion-autologging/src/AutoLogging";
 export { ALSurfaceCapability } from "@hyperion/hyperion-autologging/src/ALSurface";
 export * as ALSurfaceUtils from "@hyperion/hyperion-autologging/src/ALSurfaceUtils";
 export * as ALCustomEvent from '@hyperion/hyperion-autologging/src/ALCustomEvent';
+export { enableEventHandlers, disableEventHandlers } from "@hyperion/hyperion-autologging/src/ALUIEventPublisher";
```
reshadi, line 65: Let's `export * as ALUIEventPublisher (although publish is also exported) but at least that way the external apps are forced to use `ALUIEventPublisher.{enable|disable}UIEventHandlers.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -158,6 +158,38 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
   };
 }
 
+
+// Keep track of event handlers
+const uiEventHandlers = new Map<UIEventConfig['eventName'], {
```
reshadi, line 163: This is good, but now we had duplicate sets keeping track of event handler. It is a minor thing as of now, but could become a source of error if we don't keep these synced. May be move the new map and the two functions to the 
other module, and it already has all its public methods exported bellow. 

https://github.com/facebook/hyperion/blob/main/packages/hyperion-autologging/src/ALInteractableDOMElement.ts#L85

schwartzmx: Should I export this object,  and just have it used in ALInteractableDOMElement (just needs the map keys)?

schwartzmx:
Moved the tracking to ALInteractableDOMElement,  needed to refactor a bit (e.g. remove `trackInteractable` call) and consolidated where `installHandlers` is called

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -249,15 +276,8 @@ let installHandlers = () => {
   installHandlers = () => { }; // Done doing stuff!
 }
 
-export function isTrackedEvent(eventName: string): boolean {
-  return TrackedEvents.has(eventName);
-}
-
-export function trackInteractable(eventName: string): boolean {
```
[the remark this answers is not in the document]
schwartzmx: Install handlers is still called,  now just called in enableUIEventHandlers from publish

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -249,15 +276,8 @@ let installHandlers = () => {
...
-  installHandlers();
-  const alreadyTracked = TrackedEvents.has(eventName);
-  TrackedEvents.add(eventName);
```
schwartzmx, line 259: This mutates the set of events (potentially) and doesn't have all the handlers needed for the map/set to be combined.  

The tests still pass and the data-*able attributes are still set since install handlers is called in enableUIEventHandlers.  Open to other thoughts but the last suggestion was to combine these objects tracking events that are registered and this part was causing some complications with using the shared map when it doesn't have the necessary value object to set

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -245,43 +243,49 @@ export function publish(options: InitOptions): void {
...
+      let flowlet: IALFlowlet | undefined;
+      if (shouldPushPopFlowlet(event) && (flowlet = activeUIEventFlowlets.get(eventName)) != null) {
+        flowletManager.pop(flowlet);
+        activeUIEventFlowlets.delete(eventName);
+        uiEventFlowletManager.pop(flowlet);
+      }
+    };
+
+    // Set our handlers to register
+    UIEventHandlers.set(eventName, {
```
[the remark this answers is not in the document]
reshadi: Yes, actually I was thinking about it as well. You can add an optional argument to the `enableUIEventHandlers` function that gets the full object, and then inside of that function, you can add the item to the map if it is missing. 
That way, we a bit more duality between the `isTrackedEvent` and `enableUIEventHandler`. 

Alternatively, you can add a `trackAndEnableUIEventHandler` in the other module that has this object as required param. That way, we are more explicit about the intent, and duality of the functions names (check at the beginning and set at the end).

packages/hyperion-autologging/test/ALInteractableDOMElement.test.ts
```diff
@@ -258,7 +258,14 @@ describe("Text various element text options", () => {
       <div id="addEventListener"></div>
     `);
 
-    ALInteractableDOMElement.trackInteractable("click");
+    ALInteractableDOMElement.UIEventHandlers.set(
```
reshadi, line 261: The `trackAndEnableUIEventHandlers` suggestion above will also combine these two calls.

reshadi (approved):
Overall it looks good, but better update the function names and signature as you also suggested.

schwartzmx:
Please see updates in latest push @reshadi

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -111,7 +138,7 @@ let installHandlers = () => {
 
   IReactComponent.onReactDOMElement.add((_component, props: ReactComponentObjectProps) => {
     if (props != null) {
-      TrackedEvents.forEach(event => {
+      Array.from(UIEventHandlers.keys()).forEach(event => {
```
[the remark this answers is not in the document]
schwartzmx: Is there another way around this?  If there's a more optimal way I can investigate it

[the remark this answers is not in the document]
schwartzmx: Maybe just https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -82,15 +83,54 @@ function ignoreInteractiveElement(node: HTMLElement) {
...
+
+export function enableUIEventHandlers(eventName: UIEventConfig['eventName'], eventHandlerConfig?: TrackEventHandlerConfig | undefined): void {
+  let handlerConfig = UIEventHandlers.get(eventName) ?? eventHandlerConfig;
+  // Incoming config
+  if (eventHandlerConfig != null) {
+    // Disable the existing handlers if present, before installing the new ones
+    disableUIEventHandlers(eventName);
+    handlerConfig = eventHandlerConfig;
+  }
+  if (handlerConfig?.active === false) {
```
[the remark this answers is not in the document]
schwartzmx: Updated in latest push,  this is kind of what I did already in `trackAndEnableUIEventHandlers`,  but just updated it to push it closer to the enablement

PR #160, opened 2024-03-19
Add an option to collect default values under a surface

reshadi:
In some cases, the user may not click on any of the interactable elements and application will use the default values in such elements. This commit, adds an option to collect the value of some of these elements (radio/checkbox inputs, and selectors) and add then to the metadata of the surface that just mounted.

I also updated the test react app to ensure this is working.

### DIFF
--- a/packages/hyperion-autologging/src/ALElementValuePublisher.ts
+++ b/packages/hyperion-autologging/src/ALElementValuePublisher.ts
@@ -5,6 +5,8 @@
 'use strict';
 
 import type { Channel } from "@hyperion/hook/src/Channel";
+import * as IElement from "@hyperion/hyperion-dom/src/IElement";
+import * as IHTMLInputElement from "@hyperion/hyperion-dom/src/IHTMLInputElement";
 import * as Types from "@hyperion/hyperion-util/src/Types";
 import performanceAbsoluteNow from "@hyperion/hyperion-util/src/performanceAbsoluteNow";
 import ALElementInfo from "./ALElementInfo";
@@ -13,8 +15,9 @@ import * as ALID from './ALID';
 import { ALElementTextEvent, getElementTextEvent } from './ALInteractableDOMElement';
 import { ReactComponentData } from "./ALReactUtils";
 import type { ALChannelSurfaceEvent } from "./ALSurface";
+import { AUTO_LOGGING_SURFACE } from "./ALSurfaceConsts";
 import * as ALSurfaceMutationPublisher from "./ALSurfaceMutationPublisher";
-import { getAncestralSurfaceNode } from "./ALSurfaceUtils";
+import { getAncestralSurfaceNode, getSurfacePath } from "./ALSurfaceUtils";
 import { ALElementEvent, ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALReactElementEvent, ALSharedInitOptions } from "./ALType";
 
 export type ALElementValueEventData = Readonly<
@@ -78,7 +81,11 @@ export function publish(options: InitOptions): void {
 
   function trackElementValues(surface: string, surfaceElement: Element) {
     // The following is expensive, so try to jam everything we are interested in to the selector
-    const elements = surfaceElement.querySelectorAll<HTMLInputElement | HTMLSelectElement>('input[type=radio], input[type=checkbox], select');
+    const elements = surfaceElement.querySelectorAll<HTMLInputElement | HTMLSelectElement>(
+      'input[type=radio][checked], input[type=checkbox][checked], select:has(option[selected])'
+      // 'input[type=radio], input[type=checkbox], select'
+    );
+
 
     if (!elements.length) {
       // Nothing to add
@@ -89,6 +96,11 @@ export function publish(options: InitOptions): void {
 
     for (let i = 0; i < elements.length; ++i) {
       const element = elements[i];
+      /**
+       * The following function will look for 'interactable' surfaces, so the logic
+       * automatically filter impression only surfaces. This ensure the logic stays
+       * compatible with the UIEvent llogic.
+       */
       const parentSurfaceNode = getAncestralSurfaceNode(element);
       if (parentSurfaceNode === surfaceElement) {
         // We know value is part of this surface only
@@ -139,5 +151,45 @@ export function publish(options: InitOptions): void {
     }
 
     trackElementValues(surface, surfaceElement);
+  });
+
+  /**
+   * Sometimes in react, it first adds the empty input elements, and soon after updates the default value
+   * based on props. In these cases, we do miss the elements in the above algorithm when parent surface is
+   * mounted.
+   * Using interception, we can not only catch such initialization, but also any future change to the value.
+   * When combined with `change` ui event, we can track ALL changes to these input elements at any time.
+   * TODO: we should decide if we want to have `change` event logged as is, or to file a new value event as well.
+   */
+  IHTMLInputElement.checked.setter.onBeforeCallObserverAdd(function (this, value) {
+    if (value) {
+      const surface = getSurfacePath(this);
+      if (!surface) {
+        return;
+      }
+
+      const relatedEventIndex = -1; // Find the last ui event event_index;
+      emitEvent({
+        element: this,
+        surface,
+        relatedEventIndex,
+        value: 'true',
+        metadata: {
+          type: this.getAttribute('type') ?? '',
+        }
+      });
+    }
+  });
+
+  /**
+   * The following is an alternative way to ensure we can catch ANY surface wrapper regardless of how it is added.
+   * However, for now will keep this disabled until we first understand the performance overhead of the rest of the algorithm.
+   */
+  false && IElement.setAttribute.onBeforeCallObserverAdd(function (this, attrName, attrValue) {
+    if (attrName !== AUTO_LOGGING_SURFACE || !(this instanceof HTMLElement)) {
+      return;
+    }
+
+    trackElementValues(attrValue, this);
   });
 }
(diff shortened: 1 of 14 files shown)


packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -62,10 +64,84 @@ export type InitOptions = Types.Options<
...
   }
 >;
 
+function publishSurfaceInputDefaults(options: InitOptions): void {
+  const { channel } = options;
+
+  channel.addListener('al_surface_mutation_event', eventData => {
+    const { event, surface, element } = eventData;
+    if (event === 'mount_component' && surface != null && element != null) {
+      // The following is expensive, so try to jam everything we are interested in to the selector
```
schwartzmx, line 77: should we check capability here as well and only do this for non-impression surfaces?

reshadi: Hmm, that's a very good question. I think it does make sense to only track this for interactable surfaces, essentially whatever would have been recorded when user clicks on element. 
I'll make that change.

reshadi: Actually, this code already filters non-interactive surfaces because for each item it looks up its `parentSurfaceNode` in the dom and it can only see interactable surfaces that way. So, the impression only surface will be automatically filtered along the way.

schwartzmx: Ah yes good call

packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -62,10 +64,84 @@ export type InitOptions = Types.Options<
...
+        } | {
+          type: 'select',
+          value: string | null,
+          text: string | null,
+        }
+      } = {};
+      for (let i = 0; i < values.length; ++i) {
+        const value = values[i];
+        const parentSurfaceNode = getAncestralSurfaceNode(value);
+        if (parentSurfaceNode === element) {
```
schwartzmx, line 102: So this is checking that it's the childmost surface?

That was another question,  what happens with root surface in this case,  like `/am`?  Do we get all the input/radios etc.. on the page?

reshadi: That the goal of this code, each input is logged only under its immediate ancestor surface. So if we have something like:
`<S1>
  <I1/>
  <S2>
    <I2/>
  </S2>
</S1>`

Although both I1 and I2 are selected under the S1, only I1 will be logged. I tried a bit to play with the CSS selector to try to do this there, but could not, but anyways we can improve the selector later, but this logic ensures correctness.

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  channel.addListener('al_surface_mutation_event', surfaceEventData => {
+    const surfaceElement = surfaceEventData.element;
+    const surface = surfaceEventData.surface;
+
+    if (surfaceEventData.event !== 'mount_component' || !surface == null || !surfaceElement) {
+      return;
+    }
+    // The following is expensive, so try to jam everything we are interested in to the selector
+    const elements = surfaceElement.querySelectorAll<HTMLInputElement | HTMLSelectElement>('input[type=radio], input[type=checkbox], select');
```
schwartzmx, line 59: 😮 the generic typing

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    if (!elements.length) {
+      // Nothing to add
+      return;
+    }
+
+    for (let i = 0; i < elements.length; ++i) {
+      const element = elements[i];
+      const parentSurfaceNode = getAncestralSurfaceNode(element);
```
schwartzmx, line 68: nit but could you also note here that this under the hood will include only product surfaces?  I know you answered my earlier comment and it makes sense but at first glance and even re-reviewing I was trying to remember why this doesn't need to check capability of the surface

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -0,0 +1,131 @@
...
+      if (parentSurfaceNode === surfaceElement) {
+        // We know value is part of this surface only
+        const elementText = getElementTextEvent(element, surface);
+        let elementValueEventData: Pick<ALElementValueEventData, "value" | "metadata"> | null = null;
+
+        switch (element.nodeName) {
+          case 'INPUT': {
+            const input = element as HTMLInputElement;
```
schwartzmx, line 76: These `as` are still needed even with the generic typing of  `querySelectorAll`,  was hopeful that it could narrow the type since nodeName is most likely a static string in those types?

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -0,0 +1,131 @@
...
+            const elementInfo = ALElementInfo.getOrCreate(element);
+            reactComponentData = elementInfo.getReactComponentData();
+          }
+
+          const flowlet = options.flowletManager.top();
+
+
+          channel.emit('al_element_value_event', {
+            eventIndex: ALEventIndex.getNextEventIndex(),
+            relatedEventIndex: surfaceEventData.eventIndex,
```
schwartzmx, line 114: 👍

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -0,0 +1,131 @@
...
+          let reactComponentData: ReactComponentData | null = null;
+          if (options.cacheElementReactInfo) {
+            const elementInfo = ALElementInfo.getOrCreate(element);
+            reactComponentData = elementInfo.getReactComponentData();
+          }
+
+          const flowlet = options.flowletManager.top();
+
+
+          channel.emit('al_element_value_event', {
```
schwartzmx, line 112: Worth having an indicator in the event that it is a default state value?  I guess `element_value` is pretty telling of what it is,  just a thought

schwartzmx (approved):
Looks good overall!

packages/hyperion-autologging/src/ALElementValuePublisher.ts
reshadi: It is because the generic typing is a union of two types (`|`). Here we make it definitive. 
I wanted to avoid writing two separate `querySelectorAll` and avoid search the whole tree twice.

packages/hyperion-autologging/src/ALElementValuePublisher.ts
reshadi: I chose this name for now, wanted to go with `causeEventIndex` but thought this might be too limiting. Thoughts?

packages/hyperion-autologging/src/ALElementValuePublisher.ts
reshadi: My thought was that if in the future we tracked the changes to these values after a click, we can issue another such click.
I can add a metadata that says this is default value. 
Also, from the `relatedEventIndex` one can deduct that this was right after a mount, so much have been default value.

packages/hyperion-autologging/src/ALElementValuePublisher.ts
schwartzmx: Yeah I get that but sorry the first part was a question.  I was hopeful since you assert nodeName is INPUT above that it could automatically type narrow to this element must be an HTMLInputElement and wouldn't require the as (I don't have the code pulled up) so was just asking if the as case was necessary

packages/hyperion-autologging/src/ALElementValuePublisher.ts
schwartzmx: I think related is good,  and we can describe the scenarios for each case

packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -132,31 +130,35 @@ export function init(options: InitOptions): boolean {
...
     ALElementValuePublisher.publish({
       ...sharedOptions,
-      ...options.elementValuePublisher
+      ...options.uiEventPublisher,
+      surfaceChannel: options.surface.channel,
     });
   }
 
+  if (options.heartbeat) {
+    ALHeartbeat.start(options.heartbeat);
```
vsylee, line 158: Unrelated to this specific PR, but related to the reason you moved this: Should probably also include a comment here that heartbeat relies on UI events?

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -4,69 +4,65 @@
...
       const elementInfo = ALElementInfo.getOrCreate(element);
       reactComponentData = elementInfo.getReactComponentData();
     }
 
+    const elementText = getElementTextEvent(element, surface);
     const flowlet = options.flowletManager.top();
 
-    channel.emit('al_element_value_event', {
+    channel.emit('al_ui_event', {
+      event: 'change',
```
vsylee, line 63: Oh interesting, so you wouldn't need to enable `eventName: 'change'`  in the UI event publisher config?

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -79,14 +79,34 @@ export function publish(options: InitOptions): void {
...
+     * Some engines (e.g. Jest) may not be able to handle the more advanced
+     * query selector (e.g. :has(...)).
+     * The following code gracefully degrade to match the capabilities.
+     */
+    function tryQuery(query: string): string | null {
+      try {
+        document.querySelectorAll(query);
+        return query;
+      } catch (e) {
+        console.error('Does not handle ', query);
```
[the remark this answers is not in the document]
reshadi: No, I tied it on WWW, to make sure tests pass.

PR #162, opened 2024-03-21
Add a few missing wraped functions for flowlets

reshadi:
Both `requestIdleCallback` and `requestAnnimationFrame` have callbacks that we need to wrap to ensure flowlet continuality.

### DIFF
--- a/packages/hyperion-flowlet/src/FlowletWrappers.ts
+++ b/packages/hyperion-flowlet/src/FlowletWrappers.ts
@@ -6,7 +6,7 @@ import * as IGlobalThis from "@hyperion/hyperion-core/src/IGlobalThis";
 import * as IPromise from "@hyperion/hyperion-core/src/IPromise";
 import * as IEventTarget from "@hyperion/hyperion-dom/src/IEventTarget";
 // import * as IGlobalEventHandlers from "@hyperion/hyperion-dom/src/IGlobalEventHandlers";
-// import * as IWindow from "@hyperion/hyperion-dom/src/IWindow";
+import * as IWindow from "@hyperion/hyperion-dom/src/IWindow";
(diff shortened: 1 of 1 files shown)


packages/hyperion-flowlet/src/FlowletWrappers.ts
```diff
@@ -178,7 +178,27 @@ export function initFlowletTrackers(flowletManager: FlowletManager) {
...
+    IWindow.requestIdleCallback,
+  ]) {
+    fi.onBeforeCallMapperAdd(args => {
+      let handler = args[0];
+      args[0] = flowletManager.wrap(handler, fi.name);
+      return args;
+    });
+  }
+
+  for (const fi of [
```
vsylee, line 194: The implementation is the same, right? 

```suggestion
```

vsylee (approved):
How did you notice that these two are missing flowlet interception?

packages/hyperion-flowlet/src/FlowletWrappers.ts
reshadi: It is the same, but the types where different enough that TS was complaining. Initially I put all of them in the same block as `setTimeout` and `setInterval`. I fiddle around a bit more to see if there is a way to make TS happy with the types.

reshadi:
> How did you notice that these two are missing flowlet interception?

i was trying to setup an example for Johnny to catch tasks started from idle callback (remember your network example?). Then I noticed we have not added flowlet for these two. See the diff in the WWW

packages/hyperion-flowlet/src/FlowletWrappers.ts
reshadi: Nope, tried several things but could not make the type work.

PR #163, opened 2024-03-26
Remove customFlowlet from FlowletManager.wrap

vsylee:
The `customFlowlet` argument for `FlowletManager.wrap` is not being used anywhere. Based on flowlet design principles, we should also not be setting the custom top call flowlet when wrapping, and instead use trigger flowlet to mark custom flowlet links wherever necessary. This change removes the deprecated `customFlowlet` argument.


reshadi (approved):
Looks good. Don't forget to update the WWW types when importing this change.

PR #166, opened 2024-04-10
Refactor and break into smaller modules

reshadi:
Planning to eventually separate the bundles into as independent modules as possible. Specifically, separate the web from react, improve channel abilities (to support multi config), ...

This is just code refactor and no logic is actually changed.


schwartzmx (approved):
Lgtm

PR #168, opened 2024-04-16
Remove flowlet instances from surface publisher useLayoutEffect depen…

schwartzmx:
…dencies

These dependences seem to be causing many setup/teardowns of this effect to occur. Any time the triggerFlowlet is changing (e.g. on click) we are seeing mutation events re-occur.

This goes back to the old version prior to #160 that had only the dom* static dependencies.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -355,7 +355,7 @@ export function init(options: InitOptions): ALSurfaceHOC {
           triggerFlowlet: callFlowlet.data.triggerFlowlet
         });
       }
-    }, [domAttributeName, domAttributeValue, callFlowlet, triggerFlowlet, nodeRef]);
+    }, [domAttributeName, domAttributeValue, nodeRef]);
 
 
 


packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -355,7 +355,7 @@ export function init(options: InitOptions): ALSurfaceHOC {
           triggerFlowlet: callFlowlet.data.triggerFlowlet
         });
       }
-    }, [domAttributeName, domAttributeValue, callFlowlet, triggerFlowlet, nodeRef]);
+    }, [domAttributeName, domAttributeValue]);
```
reshadi, line 358: How about the noderRef? Shouldn't we keep that otherwise we miss new nodes added to the dom?

vsylee (approved):

vsylee:
@reshadi From https://react.dev/reference/react/useLayoutEffect#parameters:

> After every re-render with changed dependencies, React will first run the cleanup function (if you provided it) with the old values, and then run your setup function with the new values. 

This seems to be in line with @schwartzmx's findings that changes in the flowlet dependencies could cause the `al_surface_mount`/`al_surface_unmount` events to fire more often.

PR #170, opened 2024-04-17
Reduce the volume of element value change events.

reshadi:
React keeps setting the value of the input elements behind the scene (even on mouse move!) and
since we have an interception on the setter, we keep emitting the corresponding event.

This commit checks the new value of element.checked and emit event only if it has chaned.

Also, as an optimization, if the Surface does not have tracking capability, then we know that all inputs under that surface will be ignored. so, added an explicit code to filter such cases out.


schwartzmx (approved):
Changes make sense to me,  I can help test this in WWW in my morning.

Any indication of why react is reseting these values even on mouseovers?

reshadi:
> Any indication of why react is reseting these values even on mouseovers?

as I was debugging, I saw that in some validation code.

PR #171, opened 2024-04-19
Enabled multi-channel pipe as well as pausable channels.

reshadi:
In order to allow more flexibility for the applications to setup their own custom setup, and to allo more than one set of subscribers be enabled, made the following changes:
- Channel can now pipe to more than one channel. This way, we can have a core channel for main events, and mutliple application/features can make their own dedicated channels and add their own set of subscribers to those channels.
- A PausableChannel can be paused/unpaused which allows us to drop events when needed. This mechanism would have been equivalend to unsubscribe all event handlers (pause), and then add them back (unpause) when needed.


packages/hyperion-channel/test/Channel.test.ts
```diff
@@ -126,4 +126,86 @@ describe("test Channel", () => {
 
   });
 
+  test("piped multi-channels", (done) => {
+    type ChannelEventsExt = ChannelEvents & {
+      ev2: [i: boolean], // Uncomenting this line will cause
```
vsylee, line 131: Unfinished comment?

vsylee (approved):

packages/hyperion-channel/test/Channel.test.ts
reshadi: thanks for catching it. fixed.

PR #172, opened 2024-04-23
Mark default values in metadata.

reshadi:
Since we generate a `change` event for both default value (after mount) and for actual changes (e.g. after click), this commit adds logic to mark actual default values to make it easier to analyze them later.


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -151,6 +152,26 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
...
+        const input = element as HTMLInputElement;
+        value = input.checked + '';
+        metadata.type = input.getAttribute('type') ?? '';
+        break;
+      }
+      case 'SELECT': {
+        const select = element as HTMLSelectElement;
+        value = select.value;
+        metadata.type = 'select';
+        metadata.text = select.options[select.selectedIndex].text;
```
schwartzmx, line 169: Does this need to be handled by the normal text logic?

reshadi: We don't really have any select element in the app to truly test it.. Usually the <option>s of <select> are predetermined, so it should be safe. 
We can revisit this later though.

schwartzmx (approved):

PR #173, opened 2024-04-23
Expose a method to get current active UI event

reshadi:
In some cases, the application might want to add metadata to the current active ui event (in between capture and bubble). This simple PR exports a function to expose that data.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -163,6 +163,11 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
   };
 }
 
+let lastUIEvent: CurrentUIEvent | null;
+export function getCurrentUIEventData(): ALUIEventData | null | undefined {
+  return lastUIEvent?.data;
+}
+
 /**
  *
  * @param options - Configuration for determining which events to capture and emit events via provided channel.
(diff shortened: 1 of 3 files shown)


vsylee (approved):

schwartzmx (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -163,6 +163,11 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
   };
 }
 
+let lastUIEvent: CurrentUIEvent | null;
+export function getCurrentUIEventData(): ALUIEventData | null | undefined {
```
schwartzmx, line 167: I know we probably don't want to keep holding a bunch of maps around,  but thoughts on being able to find last X event?  e.g. `click`?
The map size would be small and limited to size of `uiEvents` size

reshadi: I can add that feature as well. What should the lifetime of objects in this map be? If capture->bubble, then I'm pretty sure we won't get more than one since we actively release those ourselves (even if browser would have allowed it).

An alternative is to have a weakmap that holds on to the very last event of each type and as long as someone has a reference to them, they might be alive. The danger here is that if someone updates the metadata of the event, it might be post log, but for various analysis it might be useful. 

I also feel we are approaching an implementation of event-sourcing architecture and soon might want to keep track all events with some lifetime handling mechanism. 
Now that we have a much more flexible channel piping mechanism, a more scalable approach could be a channel with memory that keeps track a buffer of events, and has a filter function that allows selection of which events to keep. 
That way, we can potentially support any type `last X event` and apply various policies to it without complicating one particular publisher. 

I can try this in a separate commit later.

PR #174, opened 2024-04-24
Add hover publisher,  configurable from UI publisher

schwartzmx:
- This adds an ability to configure `hover` from UI publisher, although there are some differences in that this publisher relies upon `mouseover` (hard reliance),  and `click` (maybe optionally if we want to relax that).

Open to suggestions on how to configure this without complicating the types/reliance on target listener api supported events.
- Ideally this is:
  - configurable as other UI events via init
  - supports additional event-specific configuration
  - could enable underlying events it depends on automatically (wish, doesn't do this yet)
  - not need to match existing target api event configuration

Example in test app:
image

### DIFF
--- a/packages/hyperion-autologging-visualizer/src/component/ALGraphInfo.react.tsx
+++ b/packages/hyperion-autologging-visualizer/src/component/ALGraphInfo.react.tsx
@@ -46,7 +46,7 @@ const StyleCss = `
 `;
 
 const DefaultOptions: ALGraph.ALGraphDynamicOptionsType = {
-  version: 1,
+  version: 2,
   events: {
     al_ui_event: {
       click: true,
--- a/packages/hyperion-autologging-visualizer/src/component/ALGraphInfo.react.tsx
+++ b/packages/hyperion-autologging-visualizer/src/component/ALGraphInfo.react.tsx
@@ -51,6 +51,7 @@ const DefaultOptions: ALGraph.ALGraphDynamicOptionsType = {
     al_ui_event: {
       click: true,
       change: false,
+      hover: false,
     },
     al_surface_mutation_event: {
       mount_component: false,
@@ -238,7 +239,7 @@ export function ALGraphInfo(props: {
 
   React.useEffect(() => {
     if (graphRef.current) {
-      // The following ensures that later the event handlers will see the latest value. 
+      // The following ensures that later the event handlers will see the latest value.
       graphRef.current.graph.setDynamicOptions(options);
     }
   }, [options]);
--- a/packages/hyperion-autologging/src/ALHoverPublisher.ts
+++ b/packages/hyperion-autologging/src/ALHoverPublisher.ts
@@ -1,0 +1,88 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+ */
+
+'use strict';
+
+import { ChannelEventType } from "@hyperion/hyperion-channel/src/Channel";
+import * as Types from "@hyperion/hyperion-util/src/Types";
+import * as ALEventIndex from "./ALEventIndex";
+import { ALSharedInitOptions } from "./ALType";
+import * as ALUIEventPublisher from "./ALUIEventPublisher";
+import { ALUIEventCaptureData } from "./ALUIEventPublisher";
+import { assert } from "@hyperion/hyperion-global";
+
+
+export type InitOptions = Types.Options<
+  Pick<ALUIEventPublisher.InitOptions, 'uiEvents'> &
+  Pick<ALSharedInitOptions<ChannelEventType<(ALUIEventPublisher.InitOptions)['channel']>>, 'channel'> & {
+    hoverDurationThresholdMS: number;
+  }
+>;
+
+export function publish(options: InitOptions): void {
+  const { channel, uiEvents, hoverDurationThresholdMS } = options;
+
+  const mouseOverConfig = uiEvents.find(e => e.eventName === 'mouseover');
+  assert(mouseOverConfig != null, 'mouseover event must be included in uiEvents config to enable hover.');
+  const clickConfig = uiEvents.find(e => e.eventName === 'click');
+  assert(clickConfig != null, 'click event must be included in uiEvents config to enable hover.');
+
+  let activeHover: ALUIEventCaptureData | null = null;
+  channel.addListener('al_ui_event_capture', eventData => {
+    if (eventData.event === 'mouseover') {
+      if (
+        // Mousing over a new target and the activeHovers targetElement
+        // is now the relatedTarget of the new event.
+        // For mouseover:
+        //    event.target (from hyperion, targetElement) – is the element where the mouse came over.
+        //    event.relatedTarget – is the element from which the mouse came (relatedTarget → target).
+        activeHover?.targetElement === eventData.domEvent?.relatedTarget
+      ) {
+        activeHover != null &&
+          maybeEmitEvent(activeHover, eventData, hoverDurationThresholdMS);
+      }
+      // Set new active hover
+      activeHover = eventData;
+    }
+    // Click should end active hover.
+    // If the click is on an interactable element that contains the hovered element, then finish
+    // the hover event, utilizing the click event timestamp
+    else if (
+      activeHover != null &&
+      eventData.event === 'click' &&
+      // Clicked element (could be exact target or interactable parent) contains the active hover element
+      eventData.element?.contains(activeHover.targetElement)
+    ) {
+      activeHover != null &&
+        maybeEmitEvent(activeHover, eventData, hoverDurationThresholdMS);
+      // Reset the active hover
+      activeHover = null;
+    }
+  });
+
+  function maybeEmitEvent(
+    startEventData: ALUIEventCaptureData,
+    leaveEventData: ALUIEventCaptureData,
+    hoverDurationThresholdMS: number,
+  ): void {
+    const hoverDuration =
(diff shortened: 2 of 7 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -176,7 +201,12 @@ export function publish(options: InitOptions): void {
   let lastUIEvent: CurrentUIEvent | null;
 
   uiEvents.forEach((eventConfig => {
-    const { eventName, cacheElementReactInfo = false } = eventConfig;
+    const { eventName } = eventConfig;
+    if (eventName === 'hover') {
+      ALHoverPublisher.publish({ channel, uiEvents, hoverDurationThresholdMS: eventConfig.hoverDurationThresholdMS ?? 0 });
```
schwartzmx, line 206: I don't particularly like this,  and I saw what we do with ElementValue publisher but this one is a bit different in that `hover` is not a legitimate eventName from `DocumentEventMap` to reuse so this causes a bit of type problems with emitting `al_ui_event`s and also reusing the typing in `UIEventConfig`.

But I'm open to attempting to make it work in that way,  or open to other thoughts.

We may consider we have other `derived` UI events in the future,  and maybe can come up with what's the best way to support them in initialization and on typing the events on the channel.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -89,7 +104,17 @@ export type UIEventConfig<T = EventHandlerMap> = {
     // Whether to cache element's react information on capture, defaults to false.
     cacheElementReactInfo?: boolean;
   }>
-}[keyof T];
+}[keyof T] | {
+  eventName: 'hover',
+  // A callable filter for this event, returning true if the event should be emitted, or false if it should be discarded up front
+  eventFilter?: (domEvent: MouseEvent) => boolean;
+  // Whether to limit to elements that are "interactable", i.e. those that have event handlers registered to the element.  Defaults to true.
+  interactableElementsOnly?: boolean;
```
schwartzmx, line 112: at the moment some of this configuration is not applied to `hover` but is required or causes type issues in many other places,  even though the code complaining about the type will never be reached the way it is initialized currently

packages/hyperion-autologging/src/ALHoverPublisher.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+>;
+
+export function publish(options: InitOptions): void {
+  const { channel, uiEvents, hoverDurationThresholdMS } = options;
+
+  const mouseOverConfig = uiEvents.find(e => e.eventName === 'mouseover');
+  assert(mouseOverConfig != null, 'mouseover event must be included in uiEvents config to enable hover.');
+  const clickConfig = uiEvents.find(e => e.eventName === 'click');
+  assert(clickConfig != null, 'click event must be included in uiEvents config to enable hover.');
```
schwartzmx, line 29: Could perhaps relax this requirement and have it be optional for `click`

packages/hyperion-autologging/src/ALUIEventPublisher.ts
schwartzmx: cc: @vsylee for thoughts on triggerFlowlet integration

packages/hyperion-react-testapp/src/AutoLoggingWrapper.ts
```diff
@@ -95,6 +95,15 @@ export function init() {
           eventName: 'change',
           cacheElementReactInfo: true,
           interactableElementsOnly: false,
+        },
+        {
+          eventName: 'mouseover',
+          cacheElementReactInfo: true,
+          interactableElementsOnly: false,
```
reshadi, line 102: One option would be to add something like `durationThresholdToEmitHoverEvent` to the `mouseover` config itself. 
Then, user can enable mouseover and if they also want separate hover event, add this one flag. 
In the hover publisher, you can then read this (just like `change` for the element value) and decide to publish or not. 
That might make the config management easier. 

We still may need a "derived" or composite event config style for future. But for now, this simplification might work.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi: we we discussed today, if you retire uiEventFlowlet, then you may not need to worry about flowlet.
In anycase, we should not special case flowlets for hover or any other ui event.

packages/hyperion-autologging/src/ALHoverPublisher.ts
```diff
@@ -0,0 +1,96 @@
...
+        maybeEmitEvent(activeHover, eventData, durationThresholdToEmitHoverEvent);
+      // Reset the active hover
+      activeHover = null;
+    }
+  });
+
+  function maybeEmitEvent(
+    startEventData: ALUIEventCaptureData,
+    leaveEventData: ALUIEventCaptureData,
+    hoverDurationThresholdMS: number,
```
reshadi, line 75: why does this need to be passed here? This function directly close on the value, right?

packages/hyperion-autologging/src/ALHoverPublisher.ts
```diff
@@ -0,0 +1,96 @@
...
+      }
+      // Set new active hover
+      activeHover = eventData;
+    }
+    // Click should end active hover.
+    // If the click is on an interactable element that contains the hovered element, then finish
+    // the hover event, utilizing the click event timestamp
+    else if (
+      activeHover != null &&
+      eventData.event === 'click' &&
```
reshadi, line 61: should we also consider 'mousedown'?

packages/hyperion-autologging/src/ALHoverPublisher.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+  });
+
+  function maybeEmitEvent(
+    startEventData: ALUIEventCaptureData,
+    leaveEventData: ALUIEventCaptureData,
+    hoverDurationThresholdMS: number,
+  ): void {
+    const hoverDuration =
+      leaveEventData.eventTimestamp - startEventData.eventTimestamp;
```
reshadi, line 78: leaving it here so it is on your radar. the eventTimestamp is the time of processing the event, not when browser fired it. If the event loop is busy, these two can be different. 
I wonder if we should generally update the `eventTImestamp` or be explicitly choose either of the times.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -224,7 +241,12 @@ export function publish(options: InitOptions): void {
 
 
   uiEvents.forEach((eventConfig => {
-    const { eventName, cacheElementReactInfo = false } = eventConfig;
+    const { eventName } = eventConfig;
+    if (eventName === 'mouseover' && eventConfig.durationThresholdToEmitHoverEvent != null) {
+      // Publish hover events as well,  if configured
+      ALHoverPublisher.publish({ channel, uiEvents });
+    }
```
reshadi, line 248: This causing circular dependency between these two modules. Also, later if we wanted to break things down into plugins, it will be harder to manager. 

It is better to just call the `ALHoverPublisher.publish({ channel, uiEvents });` in the AutoLogging.ts directly. (https://github.com/facebook/hyperion/blob/main/packages/hyperion-autologging/src/AutoLogging.ts#L156-L170). 

That way each one of these modules are both independent, but also linked to `uiEventPublisher` option.

reshadi (changes requested):
All looks good, just a few minor fixes and we should be good to go.

packages/hyperion-autologging/src/ALHoverPublisher.ts
schwartzmx: If I remove this argument and try to use the const extracted above,  it seems to think it is possibly undefined,  even though `return` would exit the entire block.
image

```
'durationThresholdToEmitHoverEvent' is possibly 'undefined'.ts(18048)
```

packages/hyperion-autologging/src/ALHoverPublisher.ts
schwartzmx: Should we be using a converted version of `eventData.domEvent.timeStamp` everywhere instead?

packages/hyperion-autologging/src/ALHoverPublisher.ts
reshadi: It is ok to keep it. 
I think the problem is because the const value is read before the check. If you check first and then assign to a const, it will know its type is not undefined.

packages/hyperion-autologging/src/ALHoverPublisher.ts
reshadi: I'm not sure about this yet. I think we can make a simple assumption when we emit the event and decide if we want to current time or `domEvent` time; then beyond that, have everyone just rely on the `eventTimestamp` just like what you did. 
So, your code is correct, just a food for thought about the timing when we emit the event.

packages/hyperion-autologging-visualizer/src/component/ALGraphInfo.react.tsx
```diff
@@ -51,6 +51,7 @@ const DefaultOptions: ALGraph.ALGraphDynamicOptionsType = {
     al_ui_event: {
       click: true,
       change: false,
+      hover: false,
```
reshadi, line 54: Increase the version number above from 1 to 2 to ensure everyone is forced to update the value.

reshadi (approved):

reshadi (approved):

PR #176, opened 2024-05-03
fix flowlet linking in visualization

schwartzmx:
With the rename this was missed,  as it's a bit dynamic because it's trying to account for varying event types,  e.g. Flowlet event has `flowlet` while everything else will have `callFlowlet` now.

image


schwartzmx:
will be included in #177

PR #178, opened 2024-05-08
Convert Surface capability to object instead of bitset

reshadi:
We used to have a bitset for surface capability. There were several problems with it:
- FlowJS does not support bit set enums, so it was hard to use it in WWW
- TrackInteraction and TracMutation were sort of redundant. If both were missing, it would have been meaningless to have a surface. We really wanted to know if a surface was interactable or not
- We will need (next commit) more complex options for capability and bit set was not supporting it.

So, this commit simplifies the option (no other logical change) to an object.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -22,9 +22,9 @@ import * as SurfaceProxy from "./ALSurfaceProxy";
 import { ALFlowletEvent, ALMetadataEvent, ALSharedInitOptions } from "./ALType";
 
 
-export type ALChannelSurfaceEventData = ALMetadataEvent & ALFlowletEvent & Readonly<{
+export type ALSurfaceEventData = ALMetadataEvent & ALFlowletEvent & Readonly<{
   surface: string;
-  element: Element | null | undefined;
+  element: Element;
   isProxy: boolean;
   capability: ALSurfaceCapability | null | undefined;
 }>;
@@ -61,8 +61,8 @@ export type ALSurfaceRenderer = (node: React.ReactNode) => React.ReactElement;
 export type ALSurfaceHOC = (props: ALSurfaceProps, renderer?: ALSurfaceRenderer) => ALSurfaceRenderer;
 
 export type ALChannelSurfaceEvent = Readonly<{
-  al_surface_mount: [ALChannelSurfaceEventData];
-  al_surface_unmount: [ALChannelSurfaceEventData];
+  al_surface_mount: [ALSurfaceEventData];
+  al_surface_unmount: [ALSurfaceEventData];
 }>;
 
 type DataType = ALFlowletDataType;
@@ -337,7 +337,7 @@ export function init(options: InitOptions): ALSurfaceHOC {
       element.setAttribute(domAttributeName, domAttributeValue);
       __DEV__ && assert(element != null, "Invalid surface effect without an element: " + surface);
 
-      const event: ALChannelSurfaceEventData = {
+      const event: ALSurfaceEventData = {
         surface: domAttributeValue,
         callFlowlet,
         triggerFlowlet,
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -26,27 +26,25 @@ export type ALChannelSurfaceEventData = ALMetadataEvent & ALFlowletEvent & Reado
   surface: string;
   element: Element | null | undefined;
   isProxy: boolean;
-  capability: ALSurfaceCapability;
+  capability: ALSurfaceCapability | null | undefined;
 }>;
 
-export enum ALSurfaceCapability {
-  TrackInteraction = 1 << 0, // Mark all interaction events with this surface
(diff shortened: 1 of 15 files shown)


packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -26,33 +26,34 @@ export type ALChannelSurfaceEventData = ALMetadataEvent & ALFlowletEvent & Reado
...
 
-const AllSurfaceCapabilityValues = [
-  ALSurfaceCapability.TrackInteraction,
-  ALSurfaceCapability.TrackMutation,
-];
-const AllSurfaceCapabilityValuesString = AllSurfaceCapabilityValues.map(c => ALSurfaceCapability[c]).join(',');
 function surfaceCapabilityToString(capability?: ALSurfaceCapability): string {
   if (!capability) {
-    return AllSurfaceCapabilityValuesString;
+    return '';
```
schwartzmx, line 48: Would default just be interactive?  Is it worth populating if it's in the default state?

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -0,0 +1,144 @@
...
+           * So, instead we have to focus on the children of the parent element
+           */
+          if (ALSurfaceUtils.isSurfaceWrapper(element)) {
+            for (let el = element.firstElementChild; el; el = el.nextElementSibling) {
+              observer.observe(el);
+            }
+          } else {
+            observer.observe(element);
+          }
+          activeSurfaces.set(element, event);
```
schwartzmx, line 67: Just trying to understand,

So here we set active surface to the parent element,  even if it's a wrapper that we are not observing?

Later this lookup,
```js
const surfaceEvent = activeSurfaces.get(element) ?? activeSurfaces.get(element.parentElement); // element or its parent, see above for .observe(...)
```
will correctly find the active surface, regardless of if it's a wrapper or one of the siblings of the parent?

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -0,0 +1,144 @@
...
+                }
+              }
+            }
+            for (const [surfaceEvent, entries] of visibleSet) {
+              let entry = entries[0];
+              __DEV__ && assert(entry != null, 'Unexpected situation');
+              if (entries.length > 1) {
+                // Need to merge the entries
+                // ??
+                console.warn("Don't know yet how to merge entries!");
```
schwartzmx, line 119: This will occur if there are multiple sibling entries that  intersect at once?

schwartzmx (approved):
Overall I think it makes sense,  just a couple questions

packages/hyperion-autologging/src/ALSurface.ts
reshadi: the reason I renamed the values is that capability only captures the features specified, so that everything that is not specified is default value as well. That's why I chose `nonInteractive` as field name as opposed to `interactive`. 

I'm not sure if we can have a good default value for all new values going forward. That's why it is left empty here. The value here would match the props of the surface.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
reshadi: Yes, the problem is that we have to use `display:content` for the style of the wrapper to make sure it does not mess up the style of its children, in other words, it has no visual presentation; and exactly because of that its visibility cannot be detected!
So, when we have wrapper, we instead add the children of the wrapper, which almost always is actually just one child. 

Later, we know what the observer we have setup will only observe surface root nodes or direct descendants. Therefore, if we find either the observed node or its parent in the map, we can proceed

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
reshadi: Yes! The logic to build this is a bit complex, and I wanted to sort out the basics first. Anyway we are using the entry value of the first child, and as of now, I don't know if anyone even cares about this value. So, wanted to postpone adding this more to later.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (!observer) {
+        observer = new IntersectionObserver(
+          (entries: IntersectionObserverEntry[], _observer: IntersectionObserver) => {
+            /**
+             * Since surface may have many children that we added above, we need to merge
+             * all the entries, however, in most cases we may have only one entry
+             */
+            const visibleSet = new Map<ALSurfaceMutationEventData, IntersectionObserverEntry[]>();
+            for (const entry of entries) {
+              if (entry.isIntersecting) {
```
reshadi, line 100: Btw, I'm filtering and opting only for when item becomes visible. There might be value in detecting when it disappear too, in which case, I can remove this condition and pass the value on the event itself. 
What do you think? I'm leaning to do the latter, but the event volume can go up if way too many surfaces turn this feature on.

schwartzmx (approved):

PR #179, opened 2024-05-14
Capture window.location.href during capture phase of UI event

schwartzmx:
If we don't capture this as early as possible then there is a chance the action caused a location change and once ultimately logged will have the URI for where the action took the user and not necessarily where the action took place.

Example,  click anything that causes a page transition.  If this lookup is delayed it will get the destination URI and not where the action took place.

Open to naming this something else,  but `windowLocationHref` seemed a bit excessive,  `uri` or `page_uri` may work so long as it's clear where it comes from.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -17,7 +17,7 @@ import { ALID, getOrSetAutoLoggingID } from "./ALID";
 import { ALElementTextEvent, TrackEventHandlerConfig, enableUIEventHandlers, getElementTextEvent, getInteractable, isTrackedEvent } from "./ALInteractableDOMElement";
 import { ReactComponentData } from "./ALReactUtils";
 import { getSurfacePath } from "./ALSurfaceUtils";
-import { ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALReactElementEvent, ALSharedInitOptions, ALTimedEvent, Metadata } from "./ALType";
+import { ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALPageEvent, ALReactElementEvent, ALSharedInitOptions, ALTimedEvent, Metadata } from "./ALType";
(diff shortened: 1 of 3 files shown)


vsylee (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -37,6 +37,8 @@ type ALUIEvent<T = EventHandlerMap> = ALTimedEvent & ALMetadataEvent & {
     isTrusted: boolean,
     // The underlying identifier assigned to this element
     autoLoggingID: ALID | null,
+    // window.location.href when the event was captured
+    href: string | null;
```
reshadi, line 41: I think `href` is more generic than page_uri since in general href can also point to anchors inside of the page, or be data or blob or ...

Isn't this what we eventually always log as `page_uri`? Should we make it more official in our types then?

schwartzmx: Yes we log it after passing it to `URI.normalize(<val>)`.  That was kept it place but definitely open to changing it here.

>Should we make it more official in our types then?
Do you mean more official in what sense?  Should we put it on all ALLoggable events?  Here was just going to add it to UI events

reshadi: "official" in a sense that we do log it in WWW, but not here. One option is to add `page_uri` to one of the core event types and log it directly in  hyperion.

schwartzmx: Should I add it to ALLoggableEvent and ensure all events emit this value?

reshadi: Hmm, I think it is better to create a `ALPageEvent` in the `ALTypes` and then add to any payload we want to. 
I'm thinking of when we want to support React Native, and `page_uri` may not have an equivalent there. 
Ideally, this should be a plugin later added for web which adds this field to all events.

reshadi (approved):

PR #181, opened 2024-05-21
Allow configuring whether to emit initial state for disabled element …

schwartzmx:
…value events

By default we only emit events for `checked=true`,  however there are some cases where the default values when disabled are useful for understanding flows.

### DIFF
--- a/packages/hyperion-autologging/src/ALElementValuePublisher.ts
+++ b/packages/hyperion-autologging/src/ALElementValuePublisher.ts
@@ -28,7 +28,7 @@ export type InitOptions = Types.Options<
 >;
 
 export function publish(options: InitOptions): void {
-  const { channel, includeInitialDisabledState = false } = options;
+  const { channel } = options;
 
   const changeEvent = options.uiEvents.find(config => config.eventName === 'change');
 
@@ -37,7 +37,12 @@ export function publish(options: InitOptions): void {
     return;
   }
 
-  const { cacheElementReactInfo } = changeEvent;
+  // Won't refine below property includeInitialDisabledState as being available without this check...
+  if (changeEvent.eventName !== 'change') {
+    return;
+  }
+
+  const { cacheElementReactInfo, includeInitialDisabledState } = changeEvent;
 
   /**
    * In most cases, the application may not have 'change' listener. We mainly want to track
--- a/packages/hyperion-autologging/src/ALElementValuePublisher.ts
+++ b/packages/hyperion-autologging/src/ALElementValuePublisher.ts
@@ -100,13 +100,24 @@ export function publish(options: InitOptions): void {
       return null;
     }
 
-    return tryQuery(
-      'input[type=radio][checked], input[type=checkbox], select:has(option[selected])'
-    ) ?? tryQuery(
-      'input[type=radio][checked], input[type=checkbox], select'
-    ) ?? (
-        'input[type=radio], input[type=checkbox], select'
-      );
+    if (includeInitialDisabledState) {
+      // Removes the [checked] for inputs when enabled
+      return tryQuery(
+        'input[type=radio][checked], input[type=checkbox], select:has(option[selected])'
+      ) ?? tryQuery(
+        'input[type=radio][checked], input[type=checkbox], select'
+      ) ?? (
+          'input[type=radio], input[type=checkbox], select'
+        );
+    } else {
+      return tryQuery(
+        'input[type=radio][checked], input[type=checkbox][checked], select:has(option[selected])'
+      ) ?? tryQuery(
+        'input[type=radio][checked], input[type=checkbox][checked], select'
+      ) ?? (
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -132,12 +136,12 @@ export function publish(options: InitOptions): void {
...
              * Also, the following check ensures we are not reading PII from other inputs, in case we expand the search
              */
-            if (input.checked) {
+            if (includeInitialDisabledState || input.checked) {
               emitEvent({
                 surface,
                 element,
                 relatedEventIndex,
-                value: 'true',
+                value: String(input.checked),
```
vsylee, line 144: Can we change the value type so that it's a boolean and not a stringified boolean?

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -132,12 +136,12 @@ export function publish(options: InitOptions): void {
              * We know we are lookging for checkbox/radio inputs, so checked is good enough
              * Also, the following check ensures we are not reading PII from other inputs, in case we expand the search
              */
-            if (input.checked) {
+            if (includeInitialDisabledState || input.checked) {
```
vsylee, line 139: `includeInitialDisabledState` would be set by app config, right? Wondering if it's better to just remove this `if` check

vsylee (approved):

packages/hyperion-autologging/src/ALElementValuePublisher.ts
reshadi: how would it work if the `includeInitialDisabledState` is false in the config?

reshadi (approved):
looks reasonable, thanks for quick fix.

schwartzmx:
As I am trying to integrate into WWW,  @reshadi  do you have any suggestions on how to pass this option?   It's sort of engrained in UIEventPublisher options:
https://github.com/facebook/hyperion/blob/c8b40d16870e5768923001f494e12d9aee4dd9f0/packages/hyperion-autologging/src/AutoLogging.ts#L154-L168

reshadi:
> As I am trying to integrate into WWW, @reshadi do you have any suggestions on how to pass this option? It's sort of engrained in UIEventPublisher options:
> 
> https://github.com/facebook/hyperion/blob/c8b40d16870e5768923001f494e12d9aee4dd9f0/packages/hyperion-autologging/src/AutoLogging.ts#L154-L168

I see the problem. I tried changing the UIEventConfig and customize for `change`, but couldn't make it work. Basically, we would have wanted this to be part of the config for `change` rather than just a new init option. 

I'm assuming this is a temporary change and we want to later remove it right? If we want to try something like that, I might be able to think of something differently.

packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -96,9 +96,9 @@ export function publish(options: InitOptions): void {
     }
 
     return tryQuery(
-      'input[type=radio][checked], input[type=checkbox][checked], select:has(option[selected])'
+      'input[type=radio][checked], input[type=checkbox], select:has(option[selected])'
```
reshadi, line 99: this might potentially worsen the performance for existing code as well. 
to be safe, you should return the old string when the flag is disabled, and have a separate one for flag enabled. 
Otherwise, even when flag is false, we do select all inputs and iterate over them.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -98,6 +98,11 @@ export type InitOptions = Types.Options<
   ALSharedInitOptions<ALChannelUIEvent> &
   {
     uiEvents: Array<UIEventConfig>;
+  } &
+  // For ALElementValuePublisher,  temp to experiment collecting default disabled state for change events
+  {
+    // Whether to enable element_value checked=false events on initial surface mount.
+    includeInitialDisabledState?: boolean;
```
reshadi, line 105: assuming this is temporary, it is ok for now, but if we had ever published the code , we could not make such changes.

schwartzmx:
@reshadi tried to play around with this a bit more,  came up with this https://github.com/facebook/hyperion/commit/f2456e7acf4c1ad23547f2a6852d495c120e39b0
Curious what you think on this approach with types and for extension

reshadi:
> @reshadi tried to play around with this a bit more, came up with this [f2456e7](https://github.com/facebook/hyperion/commit/f2456e7acf4c1ad23547f2a6852d495c120e39b0) Curious what you think on this approach with types and for extension

This looks great. If it is compiling, go for it.

schwartzmx:
ping @reshadi for re-review after type and selector change

reshadi (approved):

PR #184, opened 2024-05-31
Allow configuring whether initial mount element value events are emitted

schwartzmx:
These `change` events get emited without any action being taken which can be confusing if expecting normal `change` events.  

They are valuable,  but consumers may not want to utilize them.
They could filter metadata.is_default = false,  or disable this behavior with this config.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -97,7 +97,14 @@ type UIEventConfigMap = {
 
 // Extend UIEventConfig with additional event-specific configuration
 export type UIEventConfig = UIEventConfigMap[keyof Omit<EventHandlerMap, 'change'>]
-  | (UIEventConfigMap['change'] & { includeInitialDisabledState?: boolean });
+  | (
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -97,7 +97,14 @@ type UIEventConfigMap = {
 
 // Extend UIEventConfig with additional event-specific configuration
 export type UIEventConfig = UIEventConfigMap[keyof Omit<EventHandlerMap, 'change'>]
-  | (UIEventConfigMap['change'] & { includeInitialDisabledState?: boolean });
+  | (
+    UIEventConfigMap['change'] & {
+      // (Default: true) Whether to include default state of radio/input/select elements when surfaces are mounted.
+      includeInitialDefaultState?: boolean,
```
reshadi, line 103: why do we need to separate this from normal change event? We do fire this event with change event as well.

reshadi: never mind, I saw the WWW diff.

reshadi (approved):

PR #185, opened 2024-06-04
Enforce tryParentInteractable text extraction limits to elements with…

schwartzmx:
… installed handlers/data-*able attributes

In some cases clicking an `input` or "interactive"-type tagged element,  there may be no available text to extract as we start extraction from the targetElement itself.

Mehrdad previously added some functionality to attempt parent element if text extraction returns nothing from the targetElement.

The current implementation enforces that interactableElementsOnly setting is set to false,  to prevent walking up the dom too far.

However there are cases we are experiencing where we don't extract text when we should because of the current guard.

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -52,9 +52,11 @@ export function getInteractable(
   node: EventTarget | null,
   eventName: UIEventConfig['eventName'],
   returnInteractableNode: boolean = false,
+  // Whether to require an actual handler is assigned to determine interactiveness, rather than including "interactive" element tags
+  requireHandlerAssigned: boolean = false,
 ): HTMLElement | null {
   // https://www.w3.org/TR/2011/WD-html5-20110525/interactive-elements.html
-  const selectorString = `[${eventHandlerTrackerAttribute(eventName)}="1"],input,button,select,option,details,dialog,summary,a[href]`;
+  const selectorString = `[${eventHandlerTrackerAttribute(eventName)}="1"]${requireHandlerAssigned ? '' : ',input,button,select,option,details,dialog,summary,a[href]'}`;
   if (node instanceof HTMLElement) {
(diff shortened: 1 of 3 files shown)


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -52,9 +52,11 @@ export function getInteractable(
   node: EventTarget | null,
   eventName: UIEventConfig['eventName'],
   returnInteractableNode: boolean = false,
+  // Whether to require an actual handler is assigned to determine interactiveness, rather than including "interactive" element tags
+  requireHandlerAssigned: boolean = false,
 ): HTMLElement | null {
   // https://www.w3.org/TR/2011/WD-html5-20110525/interactive-elements.html
-  const selectorString = `[${eventHandlerTrackerAttribute(eventName)}="1"],input,button,select,option,details,dialog,summary,a[href]`;
+  const selectorString = `[${eventHandlerTrackerAttribute(eventName)}="1"]${requireHandlerAssigned ? '' : ',input,button,select,option,details,dialog,summary,a[href]'}`;
```
reshadi, line 59: This change is good. I think another problem with this code is reliance on the `eventName`. What if the app only installs 'click' handler but we need the name for `change`, `hover`, ...

I was originally also against the `data-*able=1` implementation, but I think now is the time to fix that as well. We can discuss that separately for another fix.

reshadi (approved):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
schwartzmx: Let's discuss further,  I'm curious what you are thinking of for a solution.
We have a pretty hard reliance on the data-*able attribute setting and it does cause problems for other events and where text gets extracted.  For instance for hover,  if limited to interactable (data-mouseoverable) then we would hardly trigger events since those handlers aren't installed as often as other handlers.  So when implementing that we turn the interactable enforcement off,  but then this will extract text from the targetElement,  not necessarily the interactive element.

PR #190, opened 2024-06-20
Visibility timing

reshadi:
Fixed the timing of the component visibility events. Also added them to the graph

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
@@ -8,7 +8,7 @@ import * as Types from "@hyperion/hyperion-util/src/Types";
 import performanceAbsoluteNow from "@hyperion/hyperion-util/src/performanceAbsoluteNow";
 import { ALChannelSurfaceMutationEvent, ALSurfaceMutationEventData } from "./ALSurfaceMutationPublisher";
 import * as ALSurfaceUtils from './ALSurfaceUtils';
-import { ALElementEvent, ALExtensibleEvent, ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALSharedInitOptions } from "./ALType";
+import { ALElementEvent, ALExtensibleEvent, ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALPageEvent, ALSharedInitOptions } from "./ALType";
 
 import * as ALEventIndex from './ALEventIndex';
 import { assert } from "@hyperion/hyperion-global";
@@ -20,14 +20,24 @@ export type ALSurfaceVisibilityEventData =
   ALMetadataEvent &
   ALExtensibleEvent &
   ALElementEvent &
+  ALPageEvent &
   ALLoggableEvent &
   Readonly<
     {
       surface: string;
-      event: 'component_visibility';
-      isIntersecting: boolean;
       intersectionEntry: IntersectionObserverEntry;
-    }
+    } &
+    (
+      {
+        event: 'component_visible';
+        isIntersecting: true;
+      }
+      |
+      {
+        event: 'component_hidden';
+        isIntersecting: false;
+      }
+    )
   >;
 
 export type ALChannelSurfaceVisibilityEvent = Readonly<{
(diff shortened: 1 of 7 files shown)


packages/hyperion-util/src/performanceAbsoluteNow.ts
```diff
@@ -16,12 +16,17 @@
...
+type GetTimeFunc = () => number;
+type GetTimeFuncExtensions = {
+  setFallback: (fallback: () => number) => void;
+  fromRelativeTime(timestamp: number): number;
 };
 
+let performanceAbsoluteNow: GetTimeFunc & GetTimeFuncExtensions;
+//  {
+//   (): number,
+// } & ;
```
vsylee, line 28: Delete?

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -20,14 +20,24 @@ export type ALSurfaceVisibilityEventData =
...
     {
       surface: string;
-      event: 'component_visibility';
-      isIntersecting: boolean;
       intersectionEntry: IntersectionObserverEntry;
-    }
+    } &
+    (
+      {
+        event: 'component_visibile';
```
vsylee, line 32:
```suggestion
        event: 'component_visible';
```

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -155,19 +165,26 @@ export function publish(options: InitOptions): void {
                 console.warn("Don't know yet how to merge entries!");
               }
 
+              const isIntersecting = entry.isIntersecting;
               channel.emit('al_surface_visibility_event', {
-                event: 'component_visibility',
-                eventTimestamp: performanceAbsoluteNow(),
+                ...isIntersecting
+                  ? { event: 'component_visibile', isIntersecting }
```
vsylee, line 171:
```suggestion
                  ? { event: 'component_visible', isIntersecting }
```

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -155,19 +165,26 @@ export function publish(options: InitOptions): void {
...
+                  : { event: 'component_hidden', isIntersecting },
+                // event: 'component_visibility',
```
vsylee, line 173: Delete

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -155,19 +165,26 @@ export function publish(options: InitOptions): void {
...
                 element: surfaceEvent.element,
                 autoLoggingID: surfaceEvent.autoLoggingID, // same element, same ID
-                metadata: {},
+                metadata: {
+                  emit_time: '' + performanceAbsoluteNow(), // just to keep track of the difference
+                },
                 callFlowlet: surfaceEvent.callFlowlet,
                 triggerFlowlet: surfaceEvent.triggerFlowlet,
-                isIntersecting: entry.isIntersecting,
+                // isIntersecting,
```
vsylee, line 185: Delete

packages/hyperion-util/src/performanceAbsoluteNow.ts
```diff
@@ -38,15 +43,24 @@ if (performanceIsDefined) {
...
 } else {
-  performanceAbsoluteNow = () => fallback();
+  coreFunction = () => fallback();
 }
-performanceAbsoluteNow.setFallback = setFallback;
+
+const extensions: GetTimeFuncExtensions = {
+  setFallback,
+  fromRelativeTime: (timestamp: number) => {
+    return timestamp + navigationStart;
```
vsylee, line 60: Could there be any unexpected offsets if `navigationStart` is still -1 at this point?

vsylee (approved):

packages/hyperion-util/src/performanceAbsoluteNow.ts
reshadi: I debated with myself on this one too. If `navigationStart` is still -1 at this point, then we will fallback to the Date.now and it means that the application also does not have access to `performance` and hence cannot pass relative values to the function. 
So, in practice, it should not happen. I add a warning to at least ensure this is not happening.

PR #192, opened 2024-06-24
Enable extracting text from label elements

reshadi:
In some case, there might be just a label with a 'for' attribute that describes the label of an input element. In these cases, only label points to the input, and not vice versa.
The algorithm improvement captures that as well now.


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -444,6 +444,22 @@ function getElementName(element: HTMLElement, surface: string | null, results: A
...
 
   if (!selfText) {
+    /**
+     * In some case, there might be just a label with a 'for' attribute that describes the label of an input element.
+     * In these cases, only label points to the input, and not vice versa.
+     * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#for
+     * If we find such a label, we won't need to check the inner text anymore.
+     */
+    if (element.id) {
+      const labels = document.querySelectorAll<HTMLLabelElement>(`label[for='${element.id}']`);
```
schwartzmx, line 454: I guess there's no way to limit the document or use parentElement?  Looks like they can be anywhere in the document.   I assume this querySelector should be fast since it's very limiting.

schwartzmx (approved):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
reshadi: Unfortunately that's the case. I was looking at examples in the Ads Manager and  a lot of them are not close to each other in any ways. So, the only way to further improve this is to cache the results and avoid looking for them over and over. But has to be careful to account for DOM changes in between calls.

jasvir: Just FYI (not sure of the use cases here) - this will break the selector if element.id is malformed. Also, the spec gives you element.labels which will be faster than qsa. https://www.w3.org/TR/2012/WD-html5-20121025/the-label-element.html#dom-lfe-labels

schwartzmx: Interesting,  thanks for the pointer @jasvir  cc: @reshadi 

Might be able to change this to something like this then,  given the suggestion above:
```js
    if (element instanceof HTMLInputElement) {
      const labels = element.labels;
      if (labels && labels.length > 0) {
        for (let i = 0, len = labels.length; i < len; ++i) {
          const label = labels[i];
          getTextFromInnerText({ element: label, surface }, 'label', results);
        }
        return;
      }
    }
```

reshadi: Thanks @jasvir  for the pointer. I didn't know about that api, looks like something new, right?
We actually tried to fix the issue with https://github.com/facebook/hyperion/pull/194, but will use this api instead.

jasvir: Another subtle nuance I thought of that may or may not be an issue as you try to capture this and whether it is semantics preserving: if <input id=foo type=hidden> then <label for=foo> does NOT apply. I have never read the spec with enough care to know the consequences and whether browsers comply.

It sucks that webidl hasn't really kept up with the prose changes to the spec else you could generate these kind of things automatically.

PR #194, opened 2024-06-26
Enhance label look up by ID

reshadi:
Since some element id might have some characters that breaks the css selector passed to querySelectorAll, this commit added a simple sanitization to get around that.

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -424,6 +424,63 @@ function getTextFromInnerText(domSource: ALDOMTextSource, source: ALElementText[
   return null;
 }
 
+// Some environments (like JEST) may not have CSS
+export const cssEscape = typeof window.CSS?.escape === 'function'
+  ? (str: string) => window.CSS.escape(str)
+  : (str: string) => str.replace(/['"\[\]\(\)]/g, m => {
+    return "\\" + m;
+  });
+
+//https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#labelable
+const LabelableElememts = /BUTTON|INPUT|METER|OUTPUT|PROGRESS|SELECT|TEXTAREA/;
+function getTextFromElementLabel(domSource: ALDOMTextSource, source: 'label', results: ALElementText[]): ALElementText[] | null {
+  const { element, surface } = domSource;
+
+  // Many labelable elements could have a label assigned to them a few possible ways
+  const labels = (element as { readonly labels?: NodeListOf<HTMLLabelElement>; }).labels;
+  if (labels && labels.length) {
+    for (let i = 0, len = labels.length; i < len; ++i) {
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -452,7 +452,14 @@ function getElementName(element: HTMLElement, surface: string | null, results: A
      */
     if (element.id) {
       try {
-        const labels = document.querySelectorAll<HTMLLabelElement>(`label[for='${element.id}']`);
+        // escape characters that may break the syntax of CSS selector
+        const sanitizedId = element.id.replace(/['"\[\]\(\)]/g, m => {
```
schwartzmx, line 456: Is https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static available to use or would it cover the necessary cases here?  Looks like it is specifically for use with things such as `querySelector*`

reshadi: Oh cool. I was looking for this, but landed on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape which is different and deprecated. 
I'll use the one you listed.

reshadi: one update, I noticed jest env does not have the CSS static object. So, I added logic to choose the implementation depending on availability.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -424,6 +424,63 @@ function getTextFromInnerText(domSource: ALDOMTextSource, source: ALElementText[
...
+    return "\\" + m;
+  });
+
+//https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#labelable
+const LabelableElememts = /BUTTON|INPUT|METER|OUTPUT|PROGRESS|SELECT|TEXTAREA/;
+function getTextFromElementLabel(domSource: ALDOMTextSource, source: 'label', results: ALElementText[]): ALElementText[] | null {
+  const { element, surface } = domSource;
+
+  // Many labelable elements could have a label assigned to them a few possible ways
+  const labels = (element as { readonly labels?: NodeListOf<HTMLLabelElement>; }).labels;
```
schwartzmx, line 440: This cast just removes the need for the `instanceof HTMLInputElement` check?

schwartzmx (approved):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
reshadi: Yes, although this is just a compile time check. But, I checked all the places `.labels` defined in the spec and all of them for sure are of the right type. But, since we are applying it to generic HTMLElement nodes, we have to make it explicit. 

Besides, we don't depend on the nodes being label elements since we will use their innerText anyways. So, it is safe to do compile time check only.

PR #198, opened 2024-07-11
Allow disabling elementText extraction per event.

schwartzmx:
There are some events where the elementText extraction can be potentially expensive, and when it is not needed or relevant.

Added an option to disable this on a per event basis.
- UI Events & Mutation Events
- Defaults to on for backwards compatibility

Text disabled/enabled where expected
image

In addition I want to still use the console logger for cases like this so added a dropdown in test app similar to the other mode dropdown.
image

### DIFF
--- a/packages/hyperion-autologging/src/ALElementValuePublisher.ts
+++ b/packages/hyperion-autologging/src/ALElementValuePublisher.ts
@@ -42,7 +42,7 @@ export function publish(options: InitOptions): void {
     return;
   }
 
-  const { cacheElementReactInfo, includeInitialDefaultState = true, includeInitialDefaultDisabledState = false } = changeEvent;
+  const { cacheElementReactInfo, includeInitialDefaultState = true, includeInitialDefaultDisabledState = false, enableElementTextExtraction = false } = changeEvent;
 
   /**
    * In most cases, the application may not have 'change' listener. We mainly want to track
@@ -64,7 +64,7 @@ export function publish(options: InitOptions): void {
       reactComponentData = elementInfo.getReactComponentData();
     }
 
-    const elementText = getElementTextEvent(element, surface, tryInteractiveParentTextEventName, true);
+    const elementText = enableElementTextExtraction ? getElementTextEvent(element, surface, tryInteractiveParentTextEventName, true) : getElementTextEvent(null, null);
     const callFlowlet = options.flowletManager.top();
 
     channel.emit('al_ui_event', {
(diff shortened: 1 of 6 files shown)


packages/hyperion-autologging/src/ALElementValuePublisher.ts
```diff
@@ -64,7 +64,7 @@ export function publish(options: InitOptions): void {
       reactComponentData = elementInfo.getReactComponentData();
     }
 
-    const elementText = getElementTextEvent(element, surface, tryInteractiveParentTextEventName, true);
+    const elementText = !disableElementTextExtraction ? getElementTextEvent(element, surface, tryInteractiveParentTextEventName, true) : { elementName: null, elementText: null };
```
reshadi, line 67: Ideally, the `getElementTextEvent` function itself should be reponsible to return the 'null' equivalent value. That way, we ensure it is the same all the time. 
So, we have two options:
- pass the flag directly to the function and let it decide to return something. It already has a code for `element`.
- pass a `source` or `origin`  to the function call and then only have the options `ALElementTextOptions` has this flag supported for various sources. 

I think the second options seems cleaner and doable (source would be either event names or 'surface'...

schwartzmx: Right now not passing an `element` argument to `getElementTextEvent` will return the null object as done here.  Should we do that instead?

I'm not sure I understand option 2,  ALElementTextOptions is global for all events,  how would we conditionally return something or nothing by event?

reshadi: yes, i think nulling the `element` would be better than hard coding the empty case.

Option 2 would look something like this:
* call the function with a `origin` argument which is `EventName | 'surface'`. Something like `getElementTextEvent(element, surface, ..., 'change')`
* in the `ALElementTextOptions` have a new flag like `disableTextExtractOrigins: ['surface', 'mouseover', ...]`.
* in the actual function, check the passed origin against the options. 

You don't have to do it this way, just an idea to consider.

packages/hyperion-autologging/src/ALSurfaceMutationPublisher.ts
```diff
@@ -87,14 +88,13 @@ export function publish(options: InitOptions): void {
         let info = activeSurfaces.get(surface);
         if (!info) {
           let reactComponentData: ReactComponentData | null = null;
-          let elementText: ALElementTextEvent;
+          let elementText: ALElementTextEvent = { elementName: null, elementText: null };
```
reshadi, line 91: You should not need the initialization any more here, right?

reshadi (approved):

reshadi (approved):

PR #201, opened 2024-08-07
Remove extra flowlet fields from metadata of mutation events

schwartzmx:
If these aren't being used it would be good to clean up.


schwartzmx:
Removing from logging, but keeping them in the raw events

PR #205, opened 2024-09-13
Remove the extra flowlet push that was never poped

reshadi:
As part of
https://github.com/facebook/hyperion/commit/44523f0bb5a49e4b969364902e06ffae10ede621 PR, the uiFlowlet was removed but the pop logic was still left behind. The pop function will be no op of an empty flowlet is passed to it. As a result, every ui event was pushing a flowlet that stays on the stack.
In particular, mouseover would keep pushing flowlets that mess up other cases.

This fix ensure we stick to the principle that all trigger_flowlet must be attached to the payload of the main trigger, and then interceptors of that payload (trigger) should do push/pop.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -12,7 +12,6 @@ import { TimedTrigger } from "@hyperion/hyperion-timed-trigger/src/TimedTrigger"
 import performanceAbsoluteNow from "@hyperion/hyperion-util/src/performanceAbsoluteNow";
 import ALElementInfo from './ALElementInfo';
 import * as ALEventIndex from "./ALEventIndex";
-import { IALFlowlet } from "./ALFlowletManager";
 import { ALID, getOrSetAutoLoggingID } from "./ALID";
 import { ALElementTextEvent, TrackEventHandlerConfig, enableUIEventHandlers, getElementTextEvent, getInteractable, isTrackedEvent } from "./ALInteractableDOMElement";
 import { ReactComponentData } from "./ALReactUtils";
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -289,9 +280,6 @@ export function publish(options: InitOptions): void {
       }
       flowletName += ')';
       let callFlowlet = new flowletManager.flowletCtor(flowletName, ALUIEventGroupPublisher.getGroupRootFlowlet(event));
```
schwartzmx, line 282: Sorry for missing this.

Still it is a bit unclear to me,  so triggerFlowlet is now the only flowlet that is managed?  Call flowlets are turned into trigger flowlets via `setTriggerFlowlet`?

reshadi: This was easy to miss. 
Call flowlets are the vehicle to transfer trigger flowlets around. But the right way to ensure they never accidentially overlap is to push/pop them before/after a sync function, but use a trigger-payload to carry the trigger with async calls. For that logic, you can look at `ALTriggerFlowlet.ts` and `FlowletWrappers.ts` to see how they are moved around. 

The `setTriggerFlowlet` assigns a trigger flowlet to a payload.

vsylee:
> As part of
https://github.com/facebook/hyperion/commit/44523f0bb5a49e4b969364902e06ffae10ede621 PR, the uiFlowlet was removed but the pop logic was still left behind. The pop function will be no op of an empty flowlet is passed to it.

I'm a bit confused about this statement. Isn't the code removed pushing the call flowlet (not ui flowlet) in capture phase and later popping the same call flowlet in bubble phase?

But overall I do think this change is in the right direction, I was also pretty sure that pushing mouseover call flowlets onto the stack was causing issues with correct trigger flowlet linking. Is it more that it is "safe" to remove the call flowlet push/pop here because the event handlers are already wrapped [here](https://fburl.com/wsjvpgmy) (and so call flowlet push/pop is already happening, and more correctly, at the event handler callback level)?

reshadi:
> I'm a bit confused about this statement. Isn't the code removed pushing the call flowlet (not ui flowlet) in capture phase and later popping the same call flowlet in bubble phase?

1- The pop was not working at all because the flowlet that was passed to the `.pop()` method was always undefined. So, we were leaving flowlets on the stack and the periodic cleanup was removing them all. 
2- We don't need the push pop in the function anymore because the `triggerFlowlet` is added to the event payload itself and any other event handler will get the right value pushed/popped for that handler separately. 
Note that we only need to push/pop callflowlet if we expect OUR OWN event handler to call other functions that need to use the value. But since we know interception of event handlers read the flowlet from event event payload and handles push/pop, we don't need it in this function at all. 

> But overall I do think this change is in the right direction, I was also pretty sure that pushing mouseover call flowlets onto the stack was causing issues with correct trigger flowlet linking. Is it more that it is "safe" to remove the call flowlet push/pop here because the event handlers are already wrapped [here](https://fburl.com/wsjvpgmy) (and so call flowlet push/pop is already happening, and more correctly, at the event handler callback level)?

Yes, removing this will prevent accidental/temporal association of the trigger flowlet and we should then clearly see where else the link is broken and fix it. 

Hope this is clear.

PR #206, opened 2024-09-19
Fix trigger flowlet propagation

reshadi:
Fix Trigger Flowlet logic

    A few changes in this commit:
    - Added a generic mechanism to add global flags, useful for when we want
      to add some temporary flags to test rollout of a fix/feature
    - I noticed that when we wrap the callbacks, we always delete the
      triggerFlowlet of the top of the stack. The correct logic is that we
      should only do this if we have a `getTriggerFlowlet` function and may
      change the value in every invokation based on some argument of the
      function. Otherwise we should not touch the value.
    - For UI event, we are not setting the callFlowlet correctly (based on
      how the callback was installed). This is mostly a noop since we don't
      use the callFlowlet of the UI event, but just wanted to make it more
      consistent.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -18,6 +18,7 @@ import { ReactComponentData } from "./ALReactUtils";
 import { getSurfacePath } from "./ALSurfaceUtils";
 import { ALElementEvent, ALExtensibleEvent, ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALPageEvent, ALReactElementEvent, ALSharedInitOptions, ALTimedEvent, Metadata } from "./ALType";
 import * as ALUIEventGroupPublisher from "./ALUIEventGroupPublisher";
+import * as Flags from "@hyperion/hyperion-global/src/Flags";
 
 
 /**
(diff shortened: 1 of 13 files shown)


packages/hyperion-flowlet/test/FlowletWrapper.test.ts
```diff
@@ -153,4 +153,36 @@ describe("test flow of flowlets", () => {
...
+    }, true);
+
+    div2.addEventListener("click", event => {
+      setTimeout(() => {
+        Promise.resolve().then(() => {
+          const top = flowletManager.top();
+          console.log("Top: ", top.getFullName());
+          const tf = top.data.triggerFlowlet;
+          console.log("TF: ", tf?.getFullName());
+          expect(tf).toStrictEqual(tf2);
```
vsylee, line 180: Curious, was the trigger flowlet value here different before the changes applied in this PR?

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -279,7 +280,12 @@ export function publish(options: InitOptions): void {
         flowletName += `${separator}element=${autoLoggingID}`;
       }
       flowletName += ')';
-      let callFlowlet = new flowletManager.flowletCtor(flowletName, ALUIEventGroupPublisher.getGroupRootFlowlet(event));
+
+      let triggerFlowlet = new flowletManager.flowletCtor(flowletName, ALUIEventGroupPublisher.getGroupRootFlowlet(event));
+      let callFlowlet = Flags.getFlags().preciseTriggerFlowlet
```
vsylee, line 285: Should the check be the inverse, i.e. `!Flags.getFlags().preciseTriggerFlowlet`?

packages/hyperion-flowlet/test/FlowletWrapper.test.ts
reshadi: I was hoping to create a situation like that, but it didn't happen and TF continued to work great. 
Although, after this test, I realized we probably want to intercept `dispatchEvent` api and link to the what ever cause this api call. 
That might help us distinguish between react triggered events and actual ones better. But I will add this functionality later separately.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi: I don't think so. To test and toll out the new logic I assume we set the flag to true. The "new" logic is to read `.top()` of  the stack. Otherwise, the fallback to old behavior is that callflowlet and triggerflowet will be the same. 
Note that I renamed the old `callFlowlet` to`triggerFlowlet` since that was the actual object.

vsylee: Ah makes sense, sorry I got confused

vsylee (approved):

PR #208, opened 2024-09-26
Remove clientWidth and clientHeight checks

schwartzmx:
- These checks were added a long time ago and should not be needed
- It was found that with virtualization after nodes are removed/added and this method is called it can cause layout recalculations increasing time to render


reshadi (approved):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -157,7 +179,7 @@ let installHandlers = () => {
   IReactComponent.onReactDOMElement.add((_component, props: ReactComponentObjectProps) => {
     if (props != null) {
       UIEventNames.forEach(event => {
-        if (props[SYNTHETIC_EVENT_HANDLER_MAP[event]] != null) {
+        if (props[SYNTHETIC_EVENT_HANDLER_MAP[event]] != null && !isListenerNoop(props[SYNTHETIC_EVENT_HANDLER_MAP[event]])) {
```
reshadi, line 182: read `props[SYNTHETIC_EVENT_HANDLER_MAP[event]]` into a separate variable to avoid reading it twice (this is two loads).

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -76,15 +77,34 @@ function elementHasEventHandler(node: HTMLElementWithHandlers, eventName: HTMLEl
...
-    node.tagName === 'HTML' ||
-    (node.clientHeight === innerHeight && node.clientWidth === innerWidth)
+    node.tagName === 'HTML'
   );
 }
 
+const isFunctionBodyNoop = (body: string): boolean => {
+  // Extract body between { } via regex and trim to determine noop
+  const match = body.match(/{(?<body>(.*))}/);
+  return match?.groups?.body?.trim() === '';
```
reshadi, line 89: The following should be faster and safer:
`return /^[^(]*\([^)]*\)\s*\{\s*\}/.test(body);`

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -76,15 +77,34 @@ function elementHasEventHandler(node: HTMLElementWithHandlers, eventName: HTMLEl
...
+}
+
+// Return a bool if the listener function body is a noop.
+function isListenerNoop(listener: EventListenerOrEventListenerObject): boolean | null {
+  if (typeof listener === 'function') {
```
reshadi, line 94: use `isEventListenerObject` function which is more accurate and consistent. Then use `getFunctionInterceptor(func)?.getOriginal() ?? func` as the target func.

reshadi (changes requested):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
schwartzmx: Will update,  but that doesn't work for `() => {}`.
```
> const x = () => {}
undefined
> /^[^(]*\([^)]*\)\s*\{(\s*| return; | void 0; )\}/.test((() => {}).toString())
false
> x.toString()
'() => {}'
> x()
undefined
```

reshadi: Yes, I missed that. Your updated regex seems will do it.

reshadi (approved):

packages/hyperion-flowlet/src/global.d.ts
```diff
@@ -6,4 +6,5 @@
 
 interface GlobalFlags {
   preciseTriggerFlowlet?: boolean;
-}
\ No newline at end of file
+  useNewIgnoreInteractableImpl?: boolean;
```
reshadi, line 9: Don't add this flag in this package. Add it in the `hyperion-autologging` package. The mechanism support augmenting the type globally. 
You should always assume each package being independent and separate from others.

reshadi (changes requested):

jasvir (commented):
This regex is vulnerable to denial of service attack - I can't tell if that is acceptable here but regexs of the kind xxx*yyy*zzz* results in exponential backtracking

reshadi:
> This regex is vulnerable to denial of service attack - I can't tell if that is acceptable here but regexs of the kind xxx_yyy_zzz* results in exponential backtracking

Thanks for the feedback. I thought something like `/.*x.*z/` is much slower than `/[^x]*x[^z]*z/` is that correct assumption?
Also, any suggestions for detecting empty functions?

PR #209, opened 2024-10-19
Improve performance of interactive dom element detection

reshadi:
See individual commits for more info. 
Two important changes here:
- replace `data-*able=1` attributes for each event with a single `data-interactable="|*|"` attribute that captures all events. This will allow us in future to ignore the the event name and simply know that the element has some event handler.
- Added a caching mechanism to memoize the result of interactivity detection.

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -44,9 +44,8 @@ type HTMLElementWithHandlers = HTMLElement & {
   [K in AttributeEventHandlerName]?: any;//  ((this: GlobalEventHandlers, ev: MouseEvent) => any) | null;
 }
 
-function eventHandlerTrackerAttribute(eventName: string): string {
-  return `data-${eventName}able`;
-}
+const EventHandlerTrackerAttribute = `data-interactable`;
+
 
 export function getInteractable(
   node: EventTarget | null,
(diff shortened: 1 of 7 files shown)


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -44,45 +46,113 @@ type HTMLElementWithHandlers = HTMLElement & {
...
+
+    // https://www.w3.org/TR/2011/WD-html5-20110525/interactive-elements.html
+    selectorString ??= `[${EventHandlerTrackerAttribute}*="${eventName}"]${requireHandlerAssigned ? '' : ',input,button,select,option,details,dialog,summary,a[href]'}`;
+    const element = node;
+    if ((element.matches(selectorString) || elementHasEventHandler(element, eventName as HTMLElementEventNames)) && !ignoreInteractiveElement(element)) {
+      interactable = element;
+    } else if (element.parentElement) {
+      interactable = getInteractableOptimized(element.parentElement, eventName, requireHandlerAssigned, selectorString);
+    } else {
+      interactable = null; // We also cache null to indicate we have already tried for this element. May be unsafe!
```
schwartzmx, line 80: Wouldn't this only be reached if there is no parent,  would this only be a problem for root elements potentially?

schwartzmx (approved):

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
reshadi: This is the reimplementation of the loop that we had in the previous version, but turning it into a recursive approach so that each level can check and update the cache. Always after the loop ends we were returning null, so this is going to make sure that's still the case. 
Also in the cache, I'm treating `null` as a valid value since many of the elements along the path from element to root will have root value and we want to skip processing those as well.

PR #211, opened 2024-10-26
Improve stability of ClientSessionID and EventIndex

reshadi:
We noticed that some of the event indexes at the end of a session overlap with event indexes of the next session.
We have a scheduller for PersistentData class to lazily write back updated values to storage. This scheduler may miss the final write specially if page is busy during logging and then is unloaded (reload or move to another page).

The new code does a few things to solve this:
- The timer is reduced from 100ms to 20ms. This is still better than constantly writing the value back (specially for event index)
- The scheduller is updated to handle page unload and turn go into immediate write mode.

I also updated the guid logic to generate better random data with lower overhead (T164417895).

Hopefully all of these changes reduces the chance of event index overlaps.

### DIFF
--- a/packages/hyperion-util/src/PersistentData.ts
+++ b/packages/hyperion-util/src/PersistentData.ts
@@ -4,9 +4,11 @@
 
 'use strict';
 
+import { assert } from "@hyperion/hyperion-global/src/assert";
 import { TimedTrigger } from "@hyperion/hyperion-timed-trigger/src/TimedTrigger";
 
-function getStorage(storageName: 'sessionStorage' | 'localStorage'): Storage | null {
+type IStorage = Pick<Storage, 'getItem' | 'setItem'>;
+function getStorage(storageName: 'sessionStorage' | 'localStorage'): IStorage {
   let storage: Storage;
   try {
     storage = window[storageName];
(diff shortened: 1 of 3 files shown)


packages/hyperion-util/src/PersistentData.ts
```diff
@@ -15,50 +17,73 @@ function getStorage(storageName: 'sessionStorage' | 'localStorage'): Storage | n
...
         },
         SESSION_DATA_SAVE_INTERVAL
       );
-    } else {
-      PersistentData.runner.delay(); // If not saved yet, postpone
-    }
-    PersistentData.pending.add(this);
+      if (typeof window === "object" && typeof window.addEventListener === 'function') {
+        window.addEventListener('beforeUnload', () => {
+          runner.run();
```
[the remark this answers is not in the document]
reshadi: This is right before the page unloads and there is no time to perform any more logging. May be just firing some beacon, but even that is not guaranteed: https://stackoverflow.com/questions/41909365/what-are-the-limitation-using-before-unload-event
Also, we already have heartbeat[STOP] that uses exactly this same event. So, as long as the event index next sessions are after the STOP heartbeat event, we should be good.

packages/hyperion-util/src/PersistentData.ts
```diff
@@ -15,50 +17,73 @@ function getStorage(storageName: 'sessionStorage' | 'localStorage'): Storage | n
     storage.removeItem(x);
     return storage;
   } catch (e) {
-    return null;
+    return {
```
jasvir, line 20: What's the motivation for this change?  It's weirdly expensive no?

packages/hyperion-util/src/PersistentData.ts
```diff
@@ -15,50 +17,73 @@ function getStorage(storageName: 'sessionStorage' | 'localStorage'): Storage | n
...
+          this.runner = null;
+          this.schedule = firstRun;
         },
         SESSION_DATA_SAVE_INTERVAL
       );
-    } else {
-      PersistentData.runner.delay(); // If not saved yet, postpone
-    }
-    PersistentData.pending.add(this);
+      if (typeof window === "object" && typeof window.addEventListener === 'function') {
```
jasvir, line 54: this looks fine - mostly just curious - what is the case you're defending against by checking for the presence of aEL before calling it? Is it a security boundary or can this code run in places where aEL doesn't exist?

packages/hyperion-util/src/PersistentData.ts
reshadi: Actually this to remove an unlikely null check in the expensive part. See the `save()` method bellow. 
The `gteStorage` almost always succeeds, but because of the signature, we might return null, which means later on we have to always say `this.storage?.setItem`. By not returning null, the extra null check can go away.

packages/hyperion-util/src/PersistentData.ts
reshadi: This code sits in a package that may be called outside of main thread (e.g. WebWorker). Although as of now, the storage we support are only available in the main thread, but the plan is to later also support `IndexDB`. So, Same way we check the storages, we also check if `window.*` is also available.

packages/hyperion-util/src/PersistentData.ts
jasvir: Would just returning a frozen object be overall faster? Or at least a value prop rather than a getter setter pair.

reshadi: what do you mean by getter/setter? This is the same interface that is used in the code bellow. I could turn it into a singleton, but as of now, it is called may be a few times total. so, didn't want to add the complexity for singleton.

schwartzmx (approved):

PR #212, opened 2024-10-29
Event hash plugin setup

reshadi:
Setting up the basic parts of a plugin to add event hash to each event. 
See each commit for more details.

### DIFF
--- a/packages/hyperion-autologging-plugin-eventhash/src/index.ts
+++ b/packages/hyperion-autologging-plugin-eventhash/src/index.ts
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates. All Rights Reserved.
+ */
+
+'use strict';
+import type { ALChannelEvent } from "@hyperion/hyperion-autologging/src/AutoLogging";
+import { Channel } from "@hyperion/hyperion-channel";
+import * as XXH from "xxhashjs"
+
+export function init(channel: Channel<ALChannelEvent>): void {
+  const xxh = XXH.h64(0xABCD);
+  channel.addListener('al_ui_event', eventData => {
+    const data = `${eventData.event},${eventData.pageURI.pathname},${eventData.surface ?? ''},${eventData.reactComponentName ?? ''},${eventData.elementName ?? ''}` + null;
+    eventData.metadata.hash = xxh.update(data).digest().toString(16);
+  });
+}
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -19,6 +19,7 @@ import { getSurfacePath } from "./ALSurfaceUtils";
 import { ALElementEvent, ALExtensibleEvent, ALFlowletEvent, ALLoggableEvent, ALMetadataEvent, ALPageEvent, ALReactElementEvent, ALSharedInitOptions, ALTimedEvent, Metadata } from "./ALType";
 import * as ALUIEventGroupPublisher from "./ALUIEventGroupPublisher";
 import * as Flags from "@hyperion/hyperion-global/src/Flags";
+import { getCurrMainPageUrl } from "./MainPageUrl";
 
 
 /**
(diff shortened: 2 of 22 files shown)


packages/hyperion-autologging-plugin-eventhash/src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+'use strict';
+import type { ALChannelEvent } from "@hyperion/hyperion-autologging/src/AutoLogging";
+import { Channel } from "@hyperion/hyperion-channel";
+import * as XXH from "xxhashjs"
+
+export function init(channel: Channel<ALChannelEvent>): void {
+  const xxh = XXH.h64(0xABCD);
+  channel.addListener('al_ui_event', eventData => {
+    const data = `${eventData.event},${eventData.pageURI},${eventData.surface},${eventData.reactComponentName ?? ''},${eventData.elementName}`;
```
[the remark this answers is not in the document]
reshadi: Hmm, you're right! I was expecting that for null conversation to string compiler at least generates warning! It just assumes `null` to `"null"` conversion. Will fix.

packages/hyperion-autologging-plugin-eventhash/src/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+'use strict';
+import type { ALChannelEvent } from "@hyperion/hyperion-autologging/src/AutoLogging";
+import { Channel } from "@hyperion/hyperion-channel";
+import * as XXH from "xxhashjs"
+
+export function init(channel: Channel<ALChannelEvent>): void {
+  const xxh = XXH.h64(0xABCD);
+  channel.addListener('al_ui_event', eventData => {
+    const data = `${eventData.event},${eventData.pageURI},${eventData.surface ?? ''},${eventData.reactComponentName ?? ''},${eventData.elementName ?? ''}` + null;
```
schwartzmx, line 13: One other thought I just had re-reviewing,  pageURI (full) may be quite unique because of query parameters or resource ids in URI path.  Should we consider cleaning those prior to hash?  I don't really want to add more processing on top of this but just thinking out loud.

reshadi: No, that's actually a good point. I think in the backend, we only look at the path, right?

packages/hyperion-autologging/src/MainPageUrl.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Lets make sure we are running on the main thread in the browser
+if (
+  typeof window === 'object' &&
+  typeof window.location === 'object' &&
+  typeof window.location.href === 'string'
+) {
+  // We are on the main thread and so safe to update automatically
+  // eslint-disable-next-line no-restricted-properties
+  updateMainPageUrl(window.location.href);
```
schwartzmx, line 48: Maybe I'm missing it,  but how is this continually updated?

reshadi: Yes, that's a complicated thing that I was going to address separately. Let me know how far I can push it for this commit.

reshadi: Added some basic supports that handles majority of use cases (especially with latest Chrome) and will enhance it later more.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -220,7 +221,7 @@ function getCommonEventData<T extends keyof DocumentEventMap>(eventConfig: UIEve
     autoLoggingID,
     metadata,
     value,
-    pageURI: window.location.href,
+    pageURI: getCurrMainPageUrl(),
```
schwartzmx, line 224: This is safe to land as is?  Or should we make sure we cover all cases before relying on the new tracking?

reshadi: Ok, I updated the code to cover all possible cased for change of the URL. Had to add a `window.history` interception.
Tested it on internal tools and works correctly. 
See the equivalent diff on www, all references to the `pageURI` is already fixed.

schwartzmx (approved):

PR #213, opened 2024-11-11
Updating document for hyperion

reshadi:
Setting up the structure of the documents to go deeper into each package and its capabilities.


packages/hyperion-docs/docs/intro.md
```diff
@@ -2,46 +2,48 @@
...
-
-Generate a new Docusaurus site using the **classic template**.
+## Getting Started
 
-The classic template will automatically be added to your project after you run the command:
+First install hyperion from npm
 
 ```bash
-npm init docusaurus@latest my-website classic
+npm install hyperionj
```
schwartzmx, line 25: Should be hyperionjs?

schwartzmx (approved):
Thanks for kicking this off,  will be good to keep iterating on and building out further.

packages/hyperion-docs/docs/intro.md
reshadi: Nope, that is intentional! I could not register `hyperionjs` and it was already taken. So, for now I got the `hyperionj` and should see what happens when we upload to it.

I'll land this, but will not yet upload to the website until we get to a better shape and also have a npm package landed.

PR #214, opened 2024-11-14
Npm publish

reshadi:
This is a large change but main goal is to make sure we can publish on npm.
I had to make the following changes:
- @hyperion scope requires paid account on npm and we cannot use that to publish, so I removed that from all of the packages
- hyperion-global is already registered on npm and I had to rename it to `hyperion-globals` to work.
- had to update all package.json files to prepare for npm publish.

For now, I published a version of these modules to make sure we have registered the names to ourselves. I'll make sure gradually each module is fully workable on its own.

Fun fact: I pushed the docs on the website last night and it had a reference to `hyperionj` module. Some mofo immediately registered that package https://www.npmjs.com/package/hyperionj. 

Luckily I had already registered https://www.npmjs.com/package/hyperionjs before which we can use to host all of the packages at once. 

I'll continue to test and update, but want to get early feedback from the team.


schwartzmx (approved):

reshadi:
> Build failed, does it compile locally and something we need to change in the build step as well?

Found all the issues and sent the fix. Now tests also pass.

schwartzmx (approved):

PR #217, opened 2024-12-02
Optimize trigger_flowlet for Promise.all etc.

reshadi:
The trigger flowlet logic for static methods of Promise with an array as input could become expensive for large arrays.

This change applies the following optimizations:
- If the input argument includes just one element, pass the trigger_flowlet of the input argument to the output itself.
- Instead of an array, use a Set to account for repeat trigger_flowlet values. This can happen if a lot of Promises are created from the same call stack.
- If we find more than 5 distinct trigger_flowlets, we skip the rest of them and add `...` at the end of resulting trigger flowlet to indicate so. That should put an upper bound on the cost of the logic.

### DIFF
--- a/packages/hyperion-flowlet/src/FlowletWrappers.ts
+++ b/packages/hyperion-flowlet/src/FlowletWrappers.ts
@@ -216,6 +216,8 @@ export function initFlowletTrackers(flowletManager: FlowletManager) {
     });
   }
 
+  const MAX_TRIGGER_FLOWLET_CHAIN = 5;
+
   for (const fi of [
     IPromise.all,
     IPromise.allSettled,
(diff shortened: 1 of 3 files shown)


packages/hyperion-flowlet/test/FlowletWrapper.test.ts
```diff
@@ -150,8 +150,27 @@ describe("test flow of flowlets", () => {
...
+    const p3 = Promise.resolve(2);
+
+    const all = Promise.all([p0, p1, p3]);
+    expect(getTriggerFlowlet(all)?.getFullName()).toMatch(/Promise.all\(\d+&\d+\)/); // p0 and p1 have the same TF, P3 has a new one
+
+    const race = Promise.race([p0, p1, p3]);
+    expect(getTriggerFlowlet(race)?.getFullName()).toMatch(/Promise.race\(\d+&\d+\)/);
+
+    const allSettled = Promise.allSettled([p0, p1, p2, p3]);
+    expect(getTriggerFlowlet(allSettled)?.getFullName()).toMatch(/Promise.allSettled\(\d+&\d+\)/);
```
Vasilii-Besschetnov, line 164: Noob question: I wonder why we had 3 TF before, but now getting only 2.

Vasilii-Besschetnov (commented):
Overall LGTM.
Have a question regarding test.

packages/hyperion-flowlet/src/FlowletWrappers.ts
```diff
@@ -224,21 +226,39 @@ export function initFlowletTrackers(flowletManager: FlowletManager) {
   ]) {
     fi.onBeforeAndAfterCallMapperAdd(args => {
       const iterable = args[0];
+      if (iterable.length == 1) {
```
vsylee, line 229:
```suggestion
      if (iterable.length === 1) {
```

packages/hyperion-flowlet/test/FlowletWrapper.test.ts
vsylee: p0, p1, and p2 have the same trigger flowlet, and this PR removes repeated trigger flowlet values

vsylee (approved):
Were you able to test for meaningful improvements in WWW/AM?

PR #220, opened 2025-02-14
Improve error messages and memory usage.

reshadi:
A few things in this PR:
- Improve messaging for detecting repeat surfaces
- Further improve memory by combining another Surface Map into the `ALSurfaceData`.
- added a flag to compare all recent optimizations

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -83,7 +83,8 @@ export type SurfaceComponent = (props:
      * If this is provided,  then we won't emit mutations for this surface as we are
      * doubly wrapping that surface, for surface attribution purposes.
      */
-    proxiedContext?: ALSurfaceContext.ALSurfaceContextFilledValue & {
+    proxiedContext?: {
+      mainContext: ALSurfaceContext.ALSurfaceContextFilledValue,
       container?: Element | DocumentFragment
     }
   }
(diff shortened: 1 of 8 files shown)


packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -97,11 +102,21 @@ export function publish(options: InitOptions): void {
...
+            activeSurfaces.delete(event.surface);
+          }
         }
         break;
       }
     }
 
+    if (!optimizeSurfaceMaps) {
+      // We know this is horrible! But keeping here to compare the performance of it for now
+      tmpInit();
```
schwartzmx, line 115: This is the main portion you are testing removing?

packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -300,15 +294,15 @@ export function init(options: InitOptions): ALSurfaceHOC {
...
+        surfacePath,
+        surfaceCtx,
+        nonInteractiveSurfacePath,
         callFlowlet,
+        capability,
         domAttributeName,
         domAttributeValue,
-        capability,
-      };
-      SurfaceDataMap.set(nonInteractiveSurfacePath, surfaceData);
```
schwartzmx, line 311: This is done in the constructor now?

schwartzmx (approved):
Overall looks good,  but a little hard to see what the core changes are

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
reshadi: Yep! It's crazy we missed this for so long!

packages/hyperion-autologging/src/ALSurface.ts
reshadi: Yes, I wanted to make sure there is only want to manipulate the `Map` and also have some assertion mechanism to check the health of the `Map`. It was tricky to get all cases right.

reshadi:
> Overall looks good, but a little hard to see what the core changes are

Unfortunately, some of it was part of the https://github.com/facebook/hyperion/pull/219/ PR that was later reverted. So, now I'm combining all the changes in one show on a WWW diff.

PR #226, opened 2025-02-22
Enable local options for text extraction function

reshadi:
In some cases, we might want to call the text extraction function and by pass the global options for updating or joining the individual text elements.


packages/hyperion-react-testapp/src/component/Surface.tsx
```diff
@@ -21,7 +21,7 @@ export const Surface = (props: ALSurface.ALSurfaceProps) => {
       ...props,
       capability: {
         trackVisibilityThreshold: .5,
-        trackMutation: false,
+        // trackMutation: false,
```
hankhsiao, line 24: Remove?

hankhsiao (commented):
Looks good to me.

PR #227, opened 2025-02-25
Surface metadata

reshadi:
Add support for event specific metadata on surface. 
(it may make sense to have a `catch all` metadata for events as well. )

image

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceData.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceData.ts
@@ -16,9 +16,10 @@ export type ALSurfaceEvent = Readonly<{
   surfaceData: ALSurfaceData;
 }>;
 
-export type EventMetadata = Readonly<{
+type WritableEventMetadata = {
   [eventName in keyof DocumentEventMap]?: Metadata
-}>;
+};
+export type EventMetadata = Readonly<WritableEventMetadata>;
 
 /**
  * This core class captures the general structure of the tree.
@@ -87,6 +88,8 @@ class ALSurfaceDataRoot extends ALSurfaceDataCore {
   }
 }
 
+const InheritedEventMetadataProperyName = "__uiEventMetadata";
+
 const surfacesData = new Map<string, ALSurfaceData>();
 export class ALSurfaceData extends ALSurfaceDataCore {
   static root = ALSurfaceDataRoot.singlton;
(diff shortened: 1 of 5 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -279,6 +279,10 @@ export function publish(options: InitOptions): void {
         flowletName += `${separator}surface=${surface}`;
         separator = '&';
         surfaceData = ALSurfaceData.get(surface);
+        const eventMetadata = surfaceData?.eventMetadata?.[eventName];
```
schwartzmx, line 282: Nice,  your other changes make this much easier to access now!

schwartzmx:
Is surface data the childmost surface or a combination?  Does metadata include metadata from parent and child surfaces?

reshadi:
> Is surface data the childmost surface or a combination? Does metadata include metadata from parent and child surfaces?

Oh very good question. Yes, this is the leaf node of the tree. Should we combine these all the way to the top? That might be the right thing to do.

schwartzmx:
I think if we combine them all it may make sense and be a bit easier to use - just not sure if it is expensive to do,  and just have the childmost surface metadata keys have precedence over the parent keys if there is a conflict, IMO.

reshadi:
> I think if we combine them all it may make sense and be a bit easier to use - just not sure if it is expensive to do, and just have the childmost surface metadata keys have precedence over the parent keys if there is a conflict, IMO.

This was actually complex with all possible cases, but I think the updates is handling well. Take a look.

schwartzmx (approved):
looks good!

packages/hyperion-autologging/src/ALSurfaceData.ts
```diff
@@ -152,6 +155,28 @@ export class ALSurfaceData extends ALSurfaceDataCore {
     }
     surfacesData.set(nonInteractiveSurface, this);
 
+    /**
```
jasvir, line 158: I don't know the code well so ignore this if it makes no sense.  Is this not just recreating event bubbling and event capture without handling the corner cases?  If it is, it might be a lot easier and performant to virtualize the events layer than trying to create a parallel one.

reshadi: No, this is not emitting any events. 
We have a `surface` structure in the app, the class in this module is for the nodes of that tree. It is essentially a filtered version of the dom. 

Surface is the only manual thing developers can add to further customize AutoLogging events. Customization happens via metadata. 

The change here allows one to specify what custom metadata should be added to the events that happen under this surface. 

Since surface is hierarchical, a metadata from an ancestor surface should be applied to all events under that surface as well. That's why the logic has this rather complex (and I might say myself: clever!) setup! :)

PR #228, opened 2025-02-27
Attach surface and event name to trigger flowlet data

reshadi:
We already include surface and eventName in the name of the trigger flowlet. We can also ensure this information is part of the corresponding flowlet.

This way, if a downstream process needs this data, they won't have to rely on calling `.toString` on the whole chain and then extracting the value from the string.

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -286,6 +286,12 @@ export function publish(options: InitOptions): void {
       flowletName += ')';
 
       let triggerFlowlet = new flowletManager.flowletCtor(flowletName, ALUIEventGroupPublisher.getGroupRootFlowlet(event));
+
+      if (surface) {
+        triggerFlowlet.data.surface = surface;
+        triggerFlowlet.data.triggerUIEventName = eventName;
+      }
+
       let callFlowlet = Flags.getFlags().preciseTriggerFlowlet
         ? flowletManager.top()
         : triggerFlowlet;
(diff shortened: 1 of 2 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -286,6 +286,12 @@ export function publish(options: InitOptions): void {
       flowletName += ')';
 
       let triggerFlowlet = new flowletManager.flowletCtor(flowletName, ALUIEventGroupPublisher.getGroupRootFlowlet(event));
+
+      if (surface) {
+        triggerFlowlet.data.surface = surface;
+        triggerFlowlet.data.triggerUIEventName = eventName;
+      }
+
```
vsylee, line 294: Thoughts on generalizing the implementation here to help standardize the flowlet naming standard across the board? e.g. thinking that maybe all the parameters of the flowlet name should be set in flowlet.data, and then the Flowlet constructor could take a `createName` function to construct a string name based on the flowlet.data params

vsylee (approved):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
reshadi: Yes, the typical pattern in JS is to pass either the value of type `T` or a function of type `(..)=>T` and then the constructor can check the type of this argument and use the value directly or return of the call.

An alternative to this approach is to pass an object that has a `toString()` method will be called automatically by the system itself. 

Although, this is usually less performant, and we have seen how frequently the Flowlet constructor is called. So, I'm less inclined to include that in the core Flowlet. 

However, the other Type safe way of achieving this is to sub-class the Flowlet and effectively use the new constructor function to do the logic. That is usually more effective. 

If we standardize the parameters of the `TriggerFlowlet` class, then we can extend the `Flowlet` class and then add those params, which can be used both in `.data` field and as the `.name`.

PR #231, opened 2025-03-11
Updates to support capturing events on SVGElement

schwartzmx:
Quick draft to support SVGElement in interactable checks/detection.

Since we use these assertions in many places it may be beneficial to update this to use some shared assertions.

Will put up as a draft for now and discuss further internally.

Clicking SVG in test app emits events as expected,  need to understand further how `text` should potentially be extracted.

image


packages/hyperion-autologging-visualizer/src/component/ALSessionGraph.react.tsx
```diff
@@ -35,7 +35,7 @@ type EventBody = {
   channelEventName?: string,
   callFlowlet?: IALFlowlet | null,
   copiedFlowlet?: CopiedFlowlet,
-  targetElement?: HTMLElement | null,
+  targetElement?: HTMLElement | SVGElement | null,
```
reshadi, line 38: We should simply convert the type to `Element` which is the parent type of HTML and SVG element.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -68,11 +68,11 @@ const EventHandlerTrackerAttribute = `data-interactable`;
 
 const InteractableAncestor = `interactableAncestor`;
 type InteractableAncestorCache = {
-  [index: string]: (HTMLElement | null)[];
+  [index: string]: (HTMLElement | SVGElement | null)[];
```
reshadi, line 71: In all of these cases, I think we can simply convert them to `Element`. The api we need from the object should be in Element

https://developer.mozilla.org/en-US/docs/Web/API/Element

This may also make some of the check unnecessary, since some other browser API may return `Element` and usually we should check their type.

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -240,7 +240,7 @@ let installHandlers = () => {
     event,
     _listener,
   ) {
-    if (UIEventNames.has(event) && this instanceof HTMLElement && !ignoreInteractiveElement(this)) {
+    if (UIEventNames.has(event) && (this instanceof HTMLElement || this instanceof SVGElement) && !ignoreInteractiveElement(this)) {
```
reshadi, line 243: if `Element` worked, all these checks becomes unnecessary.

packages/hyperion-react-testapp/package.json
```diff
@@ -24,10 +19,17 @@
...
+    "hyperion-globals": "*",
+    "hyperion-hook": "*",
+    "hyperion-react": "*",
+    "node": "^22.12.0",
+    "nodejs": "^0.0.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "typescript": "^5.7.3",
-    "vite": "^6.0.7",
+    "vite": "^6.2.1",
```
reshadi, line 32: I usually try to keep these updates separate from logic changes, if possible.

schwartzmx: I couldn't run the test app because of some error,  so tried upgrading
```
philschwartz@philschwartz-mbp hyperion-react-testapp % npm run start

> hyperion-react-testapp@0.3.4 start
> vite

file:///Users/philschwartz/Projects/hyperion/packages/hyperion-react-testapp/node_modules/vite/dist/node/chunks/dep-glQox-ep.js:4
import fsp, { constants as constants$9 } from 'node:fs/promises';
              ^^^^^^^^^
SyntaxError: The requested module 'node:fs/promises' does not provide an export named 'constants'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:409:24)

Node.js v18.0.0
npm ERR! Lifecycle script `start` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: hyperion-react-testapp@0.3.4 
npm ERR!   at location: /Users/philschwartz/Projects/hyperion/packages/hyperion-react-testapp 
```

packages/hyperion-autologging/src/ALInteractableDOMElement.ts
schwartzmx: Element seems to work,  I wasn't sure if we were using some specific HTMLElement property not available on Element but it seems TS does not complain and it seems to work as expected.

reshadi (approved):

PR #232, opened 2025-04-03
Surface visibility fix1

reshadi:
It is better to review the commit separately to understand the final fix.
This PR fixes two bugs in surface_visible event:
- fixes the current surface value for non-interactive surfaces
- handles multiple nested surfaces

image

### DIFF
--- a/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
@@ -53,52 +53,110 @@ export type InitOptions = Types.Options<
 export function publish(options: InitOptions): void {
   const { channel } = options;
 
+  class MapToArray<Key, Value> {
+    private map = new Map<Key, Value[]>();
+    get(key: Key): undefined | Value[] {
+      return this.map.get(key);
+    }
+    set(key: Key, value: Value) {
+      let values = this.map.get(key);
+      if (!values) {
+        values = [];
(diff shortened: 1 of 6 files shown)


packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -53,52 +53,110 @@ export type InitOptions = Types.Options<
...
-            }
-          } else {
-            observer.unobserve(element);
-            observedRoots.delete(element);
+      if (subRoots.length > 0) {
+        // we found at least one, good enough for us
+        return subRoots;
+      } else {
+        // Could not find non-surface-wrapper root, so need to go deeper
+        for (let el = element.firstElementChild; el; el = el.nextElementSibling) {
```
schwartzmx, line 109: I see so this is the case we discussed needing to account for.

This will continue just 1 level depth,  so it shouldn't be too expensive so long as immediate children isn't large

schwartzmx:
Changes look good,  will retest the scenario we found with these changes

schwartzmx (approved):

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
reshadi: It actually does recurse down, and could be deep, but assumption here is that we will quickly find something other than a surface. 
Also, even a single child along with another surface should work.

PR #235, opened 2025-04-15
[draft] Updates to VisibilityPublisher to handle Surfaces with delayed rendering of child elements

schwartzmx:
The issue we are trying to solve is if a Surface is initially rendered but it's immediate children are delayed or marked with placeholders,  we may miss initial attachment/finding/observing of viable child elements for detecting visibility changes.
This PR:
- Attaches a mutation observer on mount if no viable children are found,  it looks for additions to immediate child elements and sets up tracking once available
- Optionally (another flag) also detects removal of child nodes and updates tracking,  this is to more accurately track other (probably less common) cases where child elements we are tracking could be dynamically removed.

Added example in test app that can demonstrate the issue,  and also demonstrate the fix.


packages/hyperion-autologging/src/AutoLogging.ts
```diff
@@ -182,7 +182,7 @@ export function init(options: InitOptions): boolean {
   if (options.surfaceVisibilityPublisher) {
     ALSurfaceVisibilityPublisher.publish({
       ...sharedOptions,
-      ...options.surfaceMutationPublisher
+      ...options.surfaceVisibilityPublisher
```
reshadi, line 185: oops!

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -127,21 +135,123 @@ export function publish(options: InitOptions): void {
...
+      for (let i = 0; i < roots.length; ++i) {
+        const root = roots[i];
+        observer.observe(root);
+        observedRoots.set(root, surfaceData);
+        surfaceDataRoots.set(surfaceData, root);
+      }
+    } else if (enableDynamicChildTracking) {
+      // Special case: no viable children to observe, so observe the root element itself
+      // and set up a mutation observer to watch for child additions
+      observer.observe(element);
```
reshadi, line 150: this won't do much since element itself does not have any visible content

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -127,21 +135,123 @@ export function publish(options: InitOptions): void {
...
+              // Disconnect the mutation observer as we no longer need it
+              mutationObserver.disconnect();
+              mutationObservers.delete(element);
+            }
+          }
+        }
+      });
+
+      // Start observing for child additions
+      mutationObserver.observe(element, { childList: true });
```
reshadi, line 227: The logic of mutation observer is fixed for all elements. so we should not create new one every time. We just need one (no need for `mutationObservers`) and just decide what to `observe` or `unobserve` an element's mutations. 

I think if the subtree added one after another, you might get only of mutation event with a long nodeList, correct?
In that case, you need find the highest node in the tree and use that as the sub-tree root.
So, you need to map nodes to surfaceData, then for all nodes under a surface, find the node that contains all others ones in the list, then add that one as the root.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -127,21 +135,123 @@ export function publish(options: InitOptions): void {
...
+                  }
+                }
+              }
+            }
+            // If we are tracking removal,  we need to keep the observer around,  which could be potentially expensive
+            if (hasViableChild && !enableDynamicChildTrackingForRemoval) {
+              // Disconnect the mutation observer as we no longer need it
+              mutationObserver.disconnect();
+              mutationObservers.delete(element);
+            }
```
reshadi, line 221: If we go with a logic of have one observer and just add/remove nodes to it, do we still need to have separate flags for removal?
I think actually for correctness, we should always removal and do cleanup. 
I wonder in the cases that we do find a root element, what happens if that root element is removed. Is that a leak for us?

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
schwartzmx: Ah yes good catch,  I knew I need to add to observedRoots to get the surfacedata correct when events ultimately get emitted.  Will remove.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
schwartzmx: The concern was leaving resources lying around with just accounting for the delayed `added` case,  but given your other feedback I will take a look at using one mutation observer.

If we need to track removal too we have to keep monitoring the element's childlist and can't clean it up like this block is doing.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
schwartzmx: Let me think on how I can make it a single one,  right now there are variables scoped into the mutation observer callback,  such as the root element,  `surfaceData` and the current `observer`.

reshadi: Oh, I didn't realize it is not turning on subtree monitoring. That's actually better.
In that case, you can simplify the logic and just say, whenever the child node is added, repeat the logic of a) loop and recurse to find a suitable node, if not, b) add another mutation observer?

Basically, recursively only look at the immediate children until you find something.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
reshadi: Once something is added or removed, you should cleanup the observation. Also, when something is unmounted, you should do that. 
You can think of handling these well, and cleanup separately.
Also, I'm not sure if creating too many observers itself is cheaper or any better.

schwartzmx:
@reshadi could you review the latest changes if you have a chance?

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -119,29 +124,137 @@ export function publish(options: InitOptions): void {
...
+      surfaceData,
+    )
+    observedRoots.set(element, surfaceData);
+    surfaceDataRoots.set(surfaceData, element);
+  }
+
+  function untrackElement(observer: IntersectionObserver, element: Element, surfaceData: ALSurfaceData): void {
+    observer.unobserve(element);
+    observedRoots.delete(element, surfaceData);
+    surfaceDataRoots.delete(surfaceData, element);
```
reshadi, line 149: I think you should also remove the virtual property values, for example setting them as null.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -119,29 +124,137 @@ export function publish(options: InitOptions): void {
...
+  }
+
+  // Handle an element by either observing its suitable children or setting up a mutation observer
+  function handleElementWithChildren(
+    observer: IntersectionObserver,
+    element: Element,
+    surfaceData: ALSurfaceData
+  ): void {
```
reshadi, line 157: add `mutationObserver` as nullable as param here, see below

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -119,29 +124,137 @@ export function publish(options: InitOptions): void {
...
+              // Found suitable children, observe them
+              for (let i = 0; i < roots.length; ++i) {
+                trackElement(observer, roots[i], surfaceData);
+                break; // Only need to observe one viable child
+              }
+            } else {
+              // Still no suitable children, track the parent again
+              // Which will check for observable roots again, and attach a mutation observer if needed
+              trackElement(observer, element, surfaceData);
+            }
```
reshadi, line 206: call `handleElementWithChildren` but pass null for `mutationObserver`. This way, we ensure the same logic is applied all the time.

reshadi (approved):
Overall looks good. just a few cleanup work.

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
schwartzmx: I remember adding this `break;` in this block though which handleElementWithChildren doesn't include,  which was important I believe otherwise it would emit duplicate events.

Will have to retest it again,  it has been a bit but I believe that was needed

packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
```diff
@@ -119,29 +124,151 @@ export function publish(options: InitOptions): void {
...
+      }
+    });
+  }
+
+  // Handle an element by either observing its suitable children or setting up a mutation observer
+  function handleElementWithChildren(
+    observer: IntersectionObserver,
+    element: Element,
+    surfaceData: ALSurfaceData,
+    mutObserver: MutationObserver | null = mutationObserver,
```
reshadi, line 210: I only see two call sites to this function, and none of them pass any `mutObserver`. Did I miss something here?

schwartzmx: Yes I had started to apply your suggestion but noted that `break;` above.  Let me clean this up.

PR #237, opened 2025-04-24
[AL]Fix surface metadata logging

xiaotagn:
Context

We notice no metadata from surface wrapper log into the `ui_event_type` and the `surface` event type. We do have for the mount_event, debug and found it's because we haven't added the metadata from surface into the AL event metadata.

Repro issue: Add metadata in surface wrapper, notice metadata only log for mutation(mount) event
Screenshot 2025-04-23 at 6 09 27 PM

| before fix | after fix |
| Screenshot 2025-04-23 at 6 17 25 PM|
Screenshot 2025-04-23 at 6 19 55 PM|

### DIFF
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -314,7 +314,6 @@ export function publish(options: InitOptions): void {
         surface,
         surfaceData,
         ...elementText,
-        metadata: {...surfaceData?.metadata,...uiEventData.metadata},
         reactComponentName: reactComponentData?.name,
         reactComponentStack: reactComponentData?.stack,
       };
--- a/packages/hyperion-autologging/src/ALUIEventPublisher.ts
+++ b/packages/hyperion-autologging/src/ALUIEventPublisher.ts
@@ -314,6 +314,7 @@ export function publish(options: InitOptions): void {
         surface,
         surfaceData,
         ...elementText,
+        metadata: {...surfaceData?.metadata,...uiEventData.metadata},
         reactComponentName: reactComponentData?.name,
         reactComponentStack: reactComponentData?.stack,
       };
--- a/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
+++ b/packages/hyperion-autologging/src/ALSurfaceVisibilityPublisher.ts
@@ -85,7 +85,7 @@ export function publish(options: InitOptions): void {
     }
   }
 
-  // lookup surfaces that are mounted by their root element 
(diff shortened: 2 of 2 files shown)


packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -314,6 +314,7 @@ export function publish(options: InitOptions): void {
         surface,
         surfaceData,
         ...elementText,
+        metadata: {...uiEventData.metadata,...surfaceData?.metadata},
```
schwartzmx, line 317: Wondering what should take precedence here in this case if there is a key conflict,  should we reverse this where ui metadata takes precedence over the surface metadata?

reshadi:
The change to UI event doesn’t seem right to me.
I’ll check carefully tonight.

packages/hyperion-autologging/src/ALUIEventPublisher.ts
```diff
@@ -314,6 +314,7 @@ export function publish(options: InitOptions): void {
         surface,
         surfaceData,
         ...elementText,
+        metadata: {...surfaceData?.metadata,...uiEventData.metadata},
```
reshadi, line 317: You should not copy `surfaceData?.metadata` here. Each surface has metadata for each type of event defined which is automatically added (see above at line 282).

reshadi (changes requested):

packages/hyperion-autologging/src/ALUIEventPublisher.ts
xiaotagn: Yep, this part has issue (line 282). When I debug the missing data from UI event, the line 282 suppose to handle it but I found it's always null value from `surfaceData?.getInheriteUIEventMetadata(eventName)`. So, the `eventMetadata` never get assigned from the `surfaceData`. Then I found it's due to we haven't setup the `metadata` when the new surface event creates for the UIEvent (ALSurface.ts line 294 and line 308). Not sure how to share the code reference here. 

I will only keep the fix for `ALSurfaceVisibilityPublisher`, for the `ALUIEventPublisher.ts` will handle after utill we have a clear solution to this.

reshadi: Did you expect `<Surface metadate={...}>` to show up on the UI event, or `<Surface uiEventMetadata={...}>` ? 
The metadata of the Surface is not supposed to be added to all UI events. Only its uiEventMetadata is supposed to be added. 

Also, note that this is not as easy as just copying some fields from parent surface. It should get all the data from all ancestor surfaces to make sure data remains consistent even when new child Surfaces are added.

reshadi (approved):

PR #257, opened 2025-08-15
Example for chaining fetches

reshadi:
To test if the trigger flowlet is correct, updated the example to make a chain of calls using fetch, promise, timeout that stresses multiple paths for calculating trigger flowlet.

This already shows that the logic is not fully robust. (will address next)


packages/hyperion-flowlet/src/FlowletManager.ts
```diff
@@ -146,9 +147,56 @@ export class FlowletManager<T extends Flowlet = Flowlet> {
...
+    if (funcInterceptor.testAndSet(IS_FLOWLET_SETUP_PROP_NAME)) {
+      /**
+       * listener is already wrapped and probably has closed on another callFlowlet, or getTriggerFlowlet function
+       * so, we test here to see if we need to create a new interceptor.
+       * This is almost similar to the dependecy list of useCallback in react.
+       */
+      // 
+      const fData = funcInterceptor.getData<FlowletData>(FLOWLET_DATA_PROP_NAME);
+      assert(fData != null, `Flowlet data is not set on the interceptor! This should never happen!`);
+      if (Flags.getFlags().verboseFlowletLogs && fData.getTriggerFlowlet !== getTriggerFlowlet) {
```
schwartzmx, line 171: Super minor,  but notice we check `fData.getTriggerFlowlet !== getTriggerFlowlet` here and immediately below

reshadi:
> Read through each individual commit and overall the changes make sense, I'm curious to test this as well for better linking events

Take a look at the updated example in the network example. If you turn off the flag, and view the algraph, you'll see the broken dependency. 

`    preciseTriggerFlowlet: true,`
image


`    preciseTriggerFlowlet: false,`
image

PR #260, opened 2025-08-27
remove unused code from Surface

reshadi:
To make it easier to use surface in react native, and to enable some features by default, cleaned up old unused code.

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -4,23 +4,17 @@
 
 'use strict';
 
-import * as IElement from "hyperion-dom/src/IElement";
-import { Flowlet } from "hyperion-flowlet/src/Flowlet";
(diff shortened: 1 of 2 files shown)


packages/hyperion-react-testapp/src/AutoLoggingWrapper.ts
```diff
@@ -1,174 +1,174 @@
...
-    }
-  });
-
-  console.log('AutoLogging.init options:', AutoLogging.getInitOptions());
-  // console.log('dsid', getDomainSessionID('localhost'));
-  // console.log('dsid', getDomainSessionID());
-  console.log('sfid', getSessionFlowID());
-
-}
+/**
```
reshadi, line 1: Not sure what is this file like this!  I try to figure out why it is considered changed.

schwartzmx (approved):

PR #261, opened 2025-08-28
Add an element with a label to the list of aspects that should be ignored as the Interactive element

atootoonchianmeta:
The error is reported when an item in GeoRadioList is selected; two separate and similar events are fired. The first one is associated with the input element, and the second one is related to the label element.
The solution is to put the label element, as the list of elements should be ignored, in an interactive state.

### DIFF
--- a/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
+++ b/packages/hyperion-autologging/src/ALInteractableDOMElement.ts
@@ -148,7 +148,7 @@ let ignoreInteractiveElement: (node: Element) => boolean = node => {
    * Later, when we wanted to remove the flag, the cleanup is easy.
    */
   function ignoreInteractiveElementCore(node: Element): boolean {
-    return node.tagName === 'BODY' || node.tagName === 'HTML' ||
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALInteractableDOMElement.ts
```diff
@@ -148,7 +148,7 @@ let ignoreInteractiveElement: (node: Element) => boolean = node => {
    * Later, when we wanted to remove the flag, the cleanup is easy.
    */
   function ignoreInteractiveElementCore(node: Element): boolean {
-    return node.tagName === 'BODY' || node.tagName === 'HTML' ||
+    return node.tagName === 'BODY' || node.tagName === 'HTML' || node.tagName === 'LABEL' ||
```
reshadi, line 151: This will be very hard to remember later. Please add a comment like this:

According to mdn, a click on label, creates a click on the input. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label
We don't want to double count the click, and want to ignore the one on the lable. By not marking it interactive, we would prevent `captureHandler` to run. 

So, that means our problem. is

PR #267, opened 2025-09-09
Fix the style of the surface wrapper

atootoonchianmeta:
Using the GeoALSurface directly for GeoComponent break some UI view due to surface style. Removing  the style fix the issue

Before removing the syle
Screenshot 2025-09-09 at 10 09 51 AM

After removing the style
Screenshot 2025-09-09 at 10 11 04 AM

### DIFF
--- a/packages/hyperion-autologging/src/ALSurface.ts
+++ b/packages/hyperion-autologging/src/ALSurface.ts
@@ -46,6 +46,10 @@ export interface ALSurfaceCapability {
    * When set, will track when the provided ratio [0,1] of the surface becomes visible
    */
   trackVisibilityThreshold?: number;
+  /**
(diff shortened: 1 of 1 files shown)


packages/hyperion-autologging/src/ALSurface.ts
```diff
@@ -290,7 +290,6 @@ export function init(options: InitOptions): ALSurfaceRenderers {
         wrapperElementType,
         {
           [SURFACE_WRAPPER_ATTRIBUTE_NAME]: "1",
-          style: { display: 'contents' },
```
reshadi, line 293: So I assume as we discussed offline, you're going to add the `wrapperStyle` as a field of the capabilities, correct?

PR #269, opened 2025-09-19
Update IReactFlowlet.ts

LukeNgo12:
Optimized this code by extracting the logic into a reusable helper function to improve readability and maintainability, and added type safety with a ComponentWithFlowlet interface.

### DIFF
--- a/packages/hyperion-react/src/IReactFlowlet.ts
+++ b/packages/hyperion-react/src/IReactFlowlet.ts
@@ -88,38 +88,52 @@ export function init<
   const IS_FLOWLET_SETUP_PROP = 'isFlowletSetup';
 
   IReactComponent.onReactClassComponentIntercept.add(shadowComponent => {
-    const methods = [
-      shadowComponent.render,
-      shadowComponent.componentWillMount,
-      shadowComponent.componentDidMount,
-      shadowComponent.componentWillReceiveProps,
-      shadowComponent.shouldComponentUpdate,
-      shadowComponent.componentWillUpdate,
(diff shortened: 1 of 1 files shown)


packages/hyperion-react/src/IReactFlowlet.ts
```diff
@@ -88,38 +88,52 @@ export function init<
...
+// Replace 'any' with the actual type of your component if known.
+interface ComponentWithFlowlet extends React.Component {
+  props: {
+    // Replace with the actual type of your props.
+    [key: string]: any;
+  };
+}
+
+// A helper function to apply the flowlet logic.
+const applyFlowletInterceptor = (method: Function): void => {
```
reshadi, line 101: This is reducing the type safety of the method. The reason in the original code the whole thing was inlined was to ensure TS can type check each method type. That means in VSCode, the intelisense fully work.

packages/hyperion-react/src/IReactFlowlet.ts
```diff
@@ -88,38 +88,52 @@ export function init<
...
-      if (method.testAndSet(IS_FLOWLET_SETUP_PROP)) {
-        return;
+  // Define a type for the component with the flowlet property.
+// Replace 'any' with the actual type of your component if known.
+interface ComponentWithFlowlet extends React.Component {
+  props: {
+    // Replace with the actual type of your props.
+    [key: string]: any;
+  };
+}
```
reshadi, line 98: this type is already defined above with much stronger typing.

reshadi (changes requested):

PR #296, opened 2026-04-14
Fix WebpackModuleRuntime fallback and add React Activity support

schwartzmx:
- Fix WebpackModuleRuntime.getExports() crash when module not found in __webpack_module_cache__ (e.g. browser MCP injecting empty cache). Null-safe access gated behind `safeWebpackModuleExports` flag.

- Convert ModuleRuntime from eager IIFE to lazy getModuleRuntime() so GlobalFlags are respected. `preferMetaModuleRuntime` flag controls whether Meta's __debug runtime is checked before webpack's cache.

- Add REACT_ACTIVITY_TYPE (0xeae8) to IReactConsts, fix broken SymbolFor call, and register in visitor vtable to suppress "optimized visitors missing entry for Symbol(react.activity)" warnings.

- Upgrade react-testapp to React 19 canary with real <Activity> component test and Vite __debug exclusion.

### DIFF
--- a/packages/hyperion-core/src/IRequire.ts
+++ b/packages/hyperion-core/src/IRequire.ts
@@ -4,9 +4,10 @@
 
 import type { FunctionInterceptor, InterceptableObjectType } from './FunctionInterceptor';
 
+import "./reference";
 import { ShadowPrototype, } from './ShadowPrototype';
 import { interceptMethod } from './MethodInterceptor';
-import { assert } from 'hyperion-globals';
+import { assert, getFlags } from 'hyperion-globals';
 
 export type InterceptedModuleExports<TModuleExports extends InterceptableObjectType> = {
   [K in keyof TModuleExports]: FunctionInterceptor<TModuleExports, string, TModuleExports[K]>;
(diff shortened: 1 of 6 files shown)


packages/hyperion-core/src/IRequire.ts
```diff
@@ -73,21 +77,37 @@ class MetaModuleRuntime extends ModuleRuntimeBase {
   }
 }
 
-const ModuleRuntime: ModuleRuntimeBase = (() => {
+let _moduleRuntime: ModuleRuntimeBase | null = null;
+function getModuleRuntime(): ModuleRuntimeBase {
```
reshadi, line 81: do you expect this function to return different values on different invocations? For example, it may fail to get a _moduleRuntime for a while and then it starts working?

If the answer is no, then the previous pattern is better since it replaces the unnecessary check and is slightly faster.

reshadi (approved):

packages/hyperion-core/src/IRequire.ts
schwartzmx: I don't expect it to change, I wanted to be able to check the global flag just for added safety - which isn't set if this is invoked immediately.  When we remove the flags I can revert back.  If you think this flag isn't totally necessary I could just swap the order to prefer meta modules,  and keep the same pattern without a flag.

PR #302, opened 2026-08-12
[init] RN AutoLogging

schwartzmx:
# init: React Native AutoLogging

## Commit stack
 
- afd586d — Add safe channels and React Native JSX runtime entries

  - Adds failure-isolated hook callback invocation and typed `Channel.emitSafely()` fan-out while retaining strict `emit()` behavior.
  - Uses safe emission because AutoLogging publishes before invoking application handlers: one throwing subscriber must not block later subscribers, downstream channels, or the application handler.
  - Makes safe piping continue after listener failures and makes `PausableChannel` honor its paused state for both emission modes.
  - Adds a reusable React element-observation gate and optimized delegation to the original runtime.
  - Adds public React Native `jsx-runtime` and `jsx-dev-runtime` package entries with `jsx`, `jsxs`, `jsxDEV`, and `createElement` compatibility.
  - Exports the existing GUID helper through the supported `hyperion-util` entry.
  - Adds `ISafeEmitter` while keeping legacy `IEmitter` implementations source-compatible.

- 102fcdc — Implement committed React Native AutoLogging surfaces

  - Defines the initial React Native AutoLogging channel, configuration, event-envelope, metadata, session, screen, and heartbeat contracts.
  - Implements `ALInstrumentedElement` with stable handler wrappers, latest committed snapshots, safe event delivery, and preservation of application return values and exceptions.
  - Adds React Context-backed `ALSurface` and `ALSurfaceData` hierarchy, registry, inheritance, non-interactive ancestry, bounded serialization, and lifecycle publication.
  - Extracts portable heartbeat vocabulary, common metadata bounds, and surface-hierarchy mechanics without coupling React Native to DOM APIs.
  - Retires the original prop-mutation and render-as-mount behavior in favor of committed React lifecycle instrumentation.
  - Introduces provider, privacy, and sampling scaffolding that is removed in `5f8a3bb` when policy ownership moves to subscribers.

- 5f8a3bb — Add mobile heartbeat session and raw publishers

  - Centralizes React Native runtime setup behind `AutoLogging.init()` without requiring a React provider.
  - Removes library-owned sampling and privacy filtering so subscribers own filtering, redaction, sampling, persistence, and transport policy.
  - Publishes accepted UI labels and scalar values with source provenance and potentially-sensitive annotations.
  - Implements dependency-injected AppState heartbeat transitions, inactivity suppression, foreground/background scheduling, cleanup, and activity refresh from UI handlers.
  - Adds contiguous event indexes and in-memory application, session, and screen state with six-character base-36 identifiers.
  - Initially adds custom app events; that API and event family are subsequently removed by `20fac55`.

- afab347 — Add list deep link and React error publishers

  - Adds `useALListViewability()` with application callback composition, stable options, screen and surface context, and bounded impression deduplication.
  - Adds raw deep-link publication with explicit source validation and target URI preservation.
  - Adds React error-boundary publication for real `Error` and component-stack input without owning the application error boundary.
  - Routes these publishers through private request messages so public subscribers receive one indexed and timestamped event per accepted request.

- 7d0ee97 — Add React Native fixtures benchmarks and artifacts

  - Extracts portable AutoLogging types, heartbeat vocabulary, and surface hierarchy behavior into `hyperion-autologging-shared` for web and mobile reuse.
  - Uses a name-keyed child map for consistent web and mobile duplicate-surface behavior, including safe stale-child cleanup.
  - Rebuilds the React Native test app around public package entries with UI, surface, heartbeat, screen, list, deep-link, error, StrictMode, and Suspense fixtures.
  - Adds an application-owned event inspector with full payload and transport-envelope views, plus an `ALSurfaceData` tree inspector.
  - Adds Babel support for stable event props, runtime microbenchmarks, heap checks, and Metro/Hermes bundle-size measurements with a runnable baseline.
  - Separates web and mobile distribution commands: `npm run build` remains web-only, while `npm run build:mobile` emits the xplat handoff artifacts.
  - Adds exact web artifact-manifest validation and dedicated mobile artifact generation without source maps, Flow output, Haste loaders, fbsource dependencies, or DOM-only code in mobile Core.

- 250af9f — Make React Native JSX observation clone-safe

  - Represents application props directly on the observed outer element so `React.cloneElement()` replacements reach the original component.
  - Keeps Hyperion runtime descriptors in instrumented type closures so internal state cannot collide with or leak into application or native-view props.
  - Preserves replacement children, arbitrary props, refs, keys, receivers, and runtime arguments across `jsx`, `jsxs`, `jsxDEV`, and `createElement`.
  - Retains stable handler wrappers, latest callbacks, committed surface snapshots, return values, and thrown exceptions across rerenders.
  - Removes handler-keyed retry state so sibling controls sharing one callback retain independent labels, surfaces, and runtime snapshots.
  - Adds regressions for cloned ScrollView-style elements, injected cells, ref and prop replacement, shared handlers, StrictMode, and Suspense retries.

- 10fba0c — Restore shared interception compatibility

  - Restores the established web function-component interceptor return path so existing render and flowlet observers continue to execute.
  - Adds a web regression using the real `FunctionInterceptor`.
  - Makes safe channel piping compatible with legacy downstream `IEmitter` implementations that provide `emit()` but not `emitSafely()`.
  - Preserves strict emission and existing web event schemas and extraction behavior.

- 0736dbf — Add independent mobile AutoLogging feature gates

  - Adds typed `AutoLogging.init({features: ...})` gates for automatic UI, surface mutation, screen transition, list impression, deep-link, and React-error events.
  - Defaults omitted modern features to enabled while allowing each publisher family to be disabled independently of the runtime master switch.
  - Prevents partially disabled legacy configuration from accidentally enabling modern observation.
  - Keeps required internal state and application callbacks working when corresponding event publication is disabled.
  - Avoids installing the JSX element instrumenter when automatic UI events are disabled.
  - The custom-event gate introduced at this point is later removed with the custom-event API in `20fac55`.

- ebb64bc — Publish mobile input labels with provenance

  - Publishes explicitly supplied text-input placeholders instead of suppressing them in the core runtime.
  - Marks placeholder labels as potentially sensitive `application_text` sourced from `placeholder`.
  - Publishes accepted callback input values as potentially sensitive `user_input` with `callback_argument` provenance.
  - Leaves filtering, redaction, persistence, and transport decisions to subscribers.
  - Does not serialize raw handler argument arrays or native event objects.

- 67f5cfc — Restore legacy React Native AutoLogging compatibility

  - Retains the injected React and JSX intercepted-module initialization fields used by WWW/AMA.
  - Preserves independent legacy interception and publishing controls.
  - Restores the `al_react_component_prop` and `al_react_component_mount` compatibility channels.
  - Keeps heartbeat and modern event families disabled by default for legacy initialization.
  - Adds strict type and runtime coverage for the existing initialization fixture.
  - Adds TODOs to remove the compatibility aliases and installer after WWW/AMA migration.

- d3d2679 — Generate WWW-compatible React Native artifacts

  - Emits native-specific filenames for entries that directly import React Native or React JSX runtimes.
  - Keeps JSX observation and the legacy installer dependency-neutral under generic filenames.
  - Rewrites generated cross-artifact references to bare Haste module names.
  - Validates required entries, dependency boundaries, filenames, and import resolution during `build:mobile`.
  - Does not change normal web distribution generation.

- 6102055 — Inject React Native lifecycle dependencies

  - Removes the eager `react-native` import from the portable mobile runtime.
  - Adds narrow structural types for injected `ReactNativeModule.AppState` capabilities.
  - Accesses the injected AppState only when heartbeat is enabled.
  - Preserves disabled, heartbeat-disabled, legacy, cleanup, and idempotent behavior.
  - Emits a dependency-neutral main artifact and byte-identical `.react.native.js` alias.
  - Adds lifecycle injection, transition, cleanup, missing-adapter, and artifact dependency tests.

- 20fac55 — Use an application-owned mobile AutoLogging channel

  - Requires callers to create, subscribe to, and pass one typed channel through `AutoLogging.init()`.
  - Uses that exact channel internally without constructing or piping another public channel.
  - Removes `initALChannel`, `getALChannel`, `addChannelSubscriber`, and queued subscriber management.
  - Removes `ALAppEvent`, `logAppEvent`, `useLogAppEvent`, the custom-event feature, and the mobile custom-event publisher.
  - Keeps private runtime request messages internal while exporting only the public event contracts.
  - Updates the test app, documentation, declarations, tests, and benchmarks without changing web behavior.

- be15786 — Install React Native JSX runtimes during AutoLogging init

  - Supports the simplified modern configuration with `ReactModule`, `JSXRuntimeModule`, `JSXDevRuntimeModule`, and `ReactNativeModule`.
  - Installs supplied raw runtimes before enabling automatic UI observation so setup is atomic.
  - Runs runtime inspection and installation only when the runtime and `features.automaticUIEvents` are enabled.
  - Retains `IReactModule` and `IJsxRuntimeModule` as deprecated intercepted-module compatibility adapters.
  - Handles immutable ESM runtime namespaces safely: writable Metro or embedded facades are updated, while immutable namespaces remain unchanged and can use the package JSX entries.
  - Rolls back partially installed runtime slots when a supplied module cannot be updated.
  - Preserves explicit installer compatibility, idempotency, disabled paths, and late-initialization semantics.

- 030bf97 — Enforce React Native distribution checks in CI
  - Builds both web and mobile distributions through the normal reproducible artifact paths.
  - Runs the complete workspace test suite without conflicting Jest worker options.
  - Enforces React Native runtime timing and mounted-heap budgets.
  - Builds the production Metro baseline and fixture bundles and enforces the incremental size ceiling.
  - Leaves Hermes and simulator validation as local platform checks.

## Current initialization

```ts
AutoLogging.init({
  channel,
  react: {
    ReactModule: React,
    JSXRuntimeModule: JsxRuntime,
    JSXDevRuntimeModule: JsxDevRuntime,
    ReactNativeModule: { AppState },
  },
  features: {
    automaticUIEvents: true,
    surfaceMutationEvents: true,
    screenTransitionEvents: true,
    listImpressionEvents: true,
    deepLinkEvents: true,
    reactErrorEvents: true,
  },
});
```

Initialization must run before relevant JSX is rendered. Runtime inspection and JSX installation are skipped when the master runtime or automatic UI events are disabled. `ReactNativeModule.AppState` is accessed only when heartbeat is enabled.

## Intentional differences from web

- React Native observes supported JSX runtimes and `createElement`; web retains its established interceptor.
- React Native heartbeat uses injected AppState; web uses document visibility and window focus.
- React Native surfaces use React Context and committed registries without DOM elements.
- Both platforms share name-keyed surface hierarchy and inheritance mechanics, including duplicate-name replacement behavior.
- React Native extracts accessibility props, `testID`, placeholders, handler values, and explicit list viewability rather than DOM targets and bubbling.
- React Native session and screen state are in memory rather than browser storage.
- React Native emits accepted raw scalar data with provenance; subscribers own filtering, sampling, redaction, transport, and persistence.
- Mobile artifacts use `npm run build:mobile`; normal `npm run build` remains web-only.
- The standalone web React-element-observation artifact is excluded from `dist`; the established web implementation and behavior remain intact.

## Validation

- `npm run build:all` passed, including exact web artifact-manifest and mobile dependency checks.
- `npm test -- --ci` passed across all workspaces.
- React Native tests: 10 suites passed, 78 tests passed, and 2 benchmark tests were intentionally skipped by the ordinary test run and executed separately.
- Runtime and mounted-heap benchmarks passed all existing budgets:

  | Measurement                  |        P95/result |        Budget |
  | ---------------------------- | ----------------: | ------------: |
  | Disabled JSX                 |       0.038 µs/op |       1 µs/op |
  | Initialized disabled         |       0.041 µs/op |       1 µs/op |
  | Enabled non-interactable JSX |       0.210 µs/op |       3 µs/op |
  | Wrapper installation         |       0.056 µs/op |       8 µs/op |
  | Cloned observed element      |       1.637 µs/op |      15 µs/op |
  | Handler dispatch             |       1.292 µs/op |      15 µs/op |
  | Render/commit                | 21.478 µs/element | 60 µs/element |
  | Heap per wrapper             |       2,270 bytes |   4,096 bytes |

- Metro incremental bundle: 41,906 bytes / 50,000-byte ceiling.
- Hermes incremental bundle: 61,245 bytes / 85,000-byte ceiling.
- Targeted ESLint and Prettier checks for the changed implementation passed. Whole-repository lint retains unrelated pre-existing findings.
- Mobile artifacts contain no source maps or `sourceMappingURL` comments.
- The portable `hyperionMobileReactNative.js` artifact has no eager `react-native` or JSX-runtime import.
- `hyperionMobileReactNative.js` and `hyperionMobileReactNative.react.native.js` are byte-identical.

## Supported modern mobile public events

Every modern public payload extends:

```ts
interface ALLoggableEvent {
  readonly eventTimestamp: number; // Epoch milliseconds.
  readonly eventIndex: number; // One contiguous runtime index.
  readonly metadata: Record<string, string | number | boolean | null>;
  readonly relatedEventIndex?: number; // Present only for a real relationship.
}
```

### `al_ui_event`

```ts
interface ALUIEventData extends ALLoggableEvent {
  event: string; // click, long_press, change, focusin, focusout, submit, refresh, etc.
  sourceProp: string; // onPress, onChangeText, onRefresh, etc.
  surface?: string;
  surfaceData?: ALSurfaceDataNode;
  surfaceMetadata?: SurfaceMetadata;
  reactComponentName?: string;
  reactComponentStack?: readonly string[];
  elementText?: string;
  elementTextSource?:
    | "accessibilityLabel"
    | "aria-label"
    | "title"
    | "testID"
    | "placeholder";
  elementTextSourceType?: "developer_identifier" | "application_text";
  elementTextPotentiallySensitive?: boolean;
  elementName?: string;
  value?: unknown;
  valueSource?:
    | "callback_argument"
    | "native_event_text"
    | "element_value_prop";
  valueSourceType?: "user_input" | "control_value";
  valuePotentiallySensitive?: boolean;
  isDisabled?: true;
}
```

The runtime extracts accepted scalar values and does not serialize raw handler argument arrays or complete native event objects.

### `al_surface_mutation_event`

```ts
interface ALSurfaceMutationEventData extends ALLoggableEvent {
  event: "mount_component" | "unmount_component";
  surface: string;
  surfacePath: string;
  surfaceData: ALSurfaceDataNode;
  mountedDuration?: number; // Unmount only.
}
```

Unmount events use `relatedEventIndex` to reference their corresponding mount event.

### `al_heartbeat_event`

```ts
interface ALHeartbeatEventData extends ALLoggableEvent {
  event: "heartbeat";
  heartbeatType:
    | "START"
    | "SCHEDULED"
    | "PAGE_FOCUS_LOST"
    | "PAGE_FOCUS_GAINED"
    | "REGAIN_PAGE_VISIBILITY"
    | "STOP";
}
```

### `al_app_state_event`

```ts
interface ALAppStateEventData extends ALLoggableEvent {
  event: "app_state_change";
  appState: string;
}
```

### `al_screen_transition_event`

```ts
interface ALScreenTransitionEventData extends ALLoggableEvent {
  event: "screen_transition";
  screen: string;
  screenId: string;
  previousScreen?: string;
  previousScreenId?: string;
}
```

### `al_list_impression_event`

```ts
interface ALListImpressionEventData extends ALLoggableEvent {
  event: "list_item_visible";
  listName: string;
  itemName?: string;
  itemIndex?: number;
  surface?: string;
  surfaceMetadata?: SurfaceMetadata;
}
```

### `al_deep_link_event`

```ts
interface ALDeepLinkEventData extends ALLoggableEvent {
  event: "deep_link_open";
  source: "initial_url" | "url_event" | "notification";
  targetURI: string;
}
```

### `al_react_error_event`

```ts
interface ALReactErrorEventData extends ALLoggableEvent {
  event: "error";
  source: "react_error_boundary";
  errorName: string;
  errorMessage?: string;
  errorStack?: string;
  boundaryName?: string;
  errorCategory?: string;
  reactComponentName?: string;
  reactComponentStack?: string;
}
```

Internal `*_request` messages remain private. The application may wrap a public event for transport without changing the event shape:

```ts
interface ALTransportEnvelope<Event extends ALLoggableEvent> {
  family: string;
  event: Event;
  context: {
    appName: string;
    appSessionId: string;
    sessionId: string;
    appInstanceId: string;
    screenId: string;
    screen?: string;
  };
}
```

## Legacy compatibility events

Legacy initialization retains these channels while migration is in progress:

```ts
interface ALLegacyReactComponentPropEventData {
  component: string;
  prop: string;
  args: unknown[];
  type: "class" | "func" | "dom";
}

interface ALLegacyReactComponentMountEventData {
  surface: string;
  args: unknown[];
}
```

- `al_react_component_prop` carries `ALLegacyReactComponentPropEventData`.
- `al_react_component_mount` carries `ALLegacyReactComponentMountEventData`.
- These compatibility payloads intentionally retain their historical shapes and do not extend `ALLoggableEvent`.

## Test app changes

Interactive Buttons for testing UI events + Mutations
image

Surface Tree
image

Event Explorer
image


schwartzmx:
Closing in favor of the plugin based approach in #304

